PR #3, opened 2022-04-29
feat: CHK-356 add client nodo according NodeForPsp.wsdl

infantesimone:
### List of changes
- add models generation in maven `generate-sources` phase according `NodeForPsp.wsdl`;
- add node client to invoke `activatePaymentNotice`;
- add unit tests;


src/main/resources/application.properties
```diff
@@ -1 +1,7 @@
-
+#odo.uri=${NODO_URI}
```
giovanniberti, line 1: Is it intended to leave these commented out?

infantesimone: absolutely no, thanks!

MarcoAGreco (approved):
👍

PR #23, opened 2022-05-26
[CHK-427] Add DDD-style `Transaction` aggregate

giovanniberti:
#### List of Changes

* Add DDD-style `Transaction` aggregate

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/model/Transaction.java
```diff
@@ -0,0 +1,35 @@
+package it.pagopa.transactions.model;
```
infantesimone, line 1: what do you think if we put all the `aggregate` and `valueObjet` in a different package? like domain?

giovanniberti: I'm ok with that, would it be everything besides `SoapEnvelope`? If it boils down to that we could even rename `model` → `domain` and put `SoapEnvelope` in some other package (just saying, I don't feel like `SoapEnvelope` belongs to "model")?

infantesimone: correct!

infantesimone:
solved the coverage and some sonar issues, for me it's ok 🚀

MarcoAGreco (dismissed):

PR #48, opened 2022-09-16
[CHK-630] Add mechanism to get transaction data by replaying events

giovanniberti:
#### Motivation and Context

This PR sets up a mechanism to get a transaction aggregate with all collected data so far from a stream of events.
The ideal target API to get such an aggregate would be something like:
```java
Transaction replayAllEvents(Stream<AbstractTransactionEvent> events, Transaction baseTransaction) {
    return events.reduce(baseTransaction, (t, e) -> t.applyEvent(e));
}
```
Also, it would also be nice to provide a way to pattern-match/downcast a `Transaction` to a more specific class to get relevant data in order to be both specific and generic as needed.
One consideration to be taken care of when designing this mechanism is that attributes in the aggregate are in a matrioska-doll, meaning that the aggregate only expands as events are applied to it.

The approach I used here is to:
 * Associate each set of added attributes to an abstract POJO `P`, where given two events `e1` and `e2` if `e2 > e1` (e2 comes after e1) then `P2 extends P1`
 * Associate each POJO to an aggregate class that implements the `Transaction` sealed interface. Each class extends the associated POJO thus inheriting all its attributes
 * Have an interface, `EventUpdatable<TargetTransactionType, EventType>` to provide operations to convert between aggregate classes by applying events (note that by using two generic arguments it's not easily possible to provide two applications `e1 → t1` and `e2 → t2` for the same aggregate class)
 * Polymorphic event application is managed through a generic method on the `Transaction` interface. Implementations are required to check for concrete types and forward the arguements to the `EventUpdateable` implementation.

One downside of this approach as currently implemented is that getting an attribute from the aggregate requires downcasting to the correct aggregate class(*) **and** getting the relevant data from the event data, e.g.:

```java
Transaction input = /* ... */
TransactionClosed t = (TransactionClosed) input;
AuthorizationResultDto authResult = t.getTransactionAuthorizationStatusUpdateData().getAuthorizationResult();
```

This is because currently POJOs directly embed event data instead of copy-pasting fields, and could be worked around if needed.

(*)**NOTE:** Casting to superclasses is fine. Thus if `A2` can be derived by applying an event to `A1` and I want fields in `A1` I can just cast an aggregate to `A1` without problems.

##### Why POJOs and not just aggregate classes?

Consider this example:
* Aggregate class A1 implements `EventUpdatable<E1, A2>` (i.e. by applying event E1 to A1 we get A2)
* Aggregate class A2 inherits from A1 (to inherit all its fields)

With this scheme we get that A2 also implements `EventUpdatable<E1, A2>` (which could make no sense?). Also, we can't get A2 to implement `EventUpdatable<E2, A3>` for example.
POJOs are a way to disentangle inheriting fields from implementing transformation through events.

#### Types of changes

#### Checklist:


giovanniberti:
This PR doesn't provide a concrete mechanism right now, before implementing one I'd like a review on the general approach and design proposed here 😃

infantesimone (commented):
great work!
2 considerations:
- I'm not sure the real benefits of having a more specific class.
- test coverage

giovanniberti:
> I'm not sure the real benefits of having a more specific class.

What class in particular?

infantesimone:
`AggragateRoot` has a status, and given a `idTransaction` and a list of events the aggragate is rebuilt. The various specialized aggregates simplifies the code?

src/main/java/it/pagopa/transactions/domain/TransactionWithRequestedAuthorization.java
```diff
@@ -0,0 +1,28 @@
...
+import it.pagopa.transactions.domain.pojos.BaseTransactionWithRequestedAuthorization;
+
+@AggregateRoot
+public final class TransactionWithRequestedAuthorization extends BaseTransactionWithRequestedAuthorization implements EventUpdatable<TransactionWithCompletedAuthorization, TransactionAuthorizationStatusUpdatedEvent>, Transaction {
+    TransactionWithRequestedAuthorization(BaseTransaction transaction, TransactionAuthorizationRequestedEvent event) {
+        super(transaction, event.getData());
+    }
+
+    @Override
+    public TransactionWithCompletedAuthorization apply(TransactionAuthorizationStatusUpdatedEvent event) {
```
infantesimone, line 16: is new aggregate?

giovanniberti: Yes, each of `TransactionWith*` is a new aggregate. An alternative would have been to make the aggregate class generic and make it hold a `T data` field, like we do with events, but I felt separate classes were cleaner here. 🙂

infantesimone (approved):

PR #57, opened 2022-10-12
[CHK-676] Add AI to Ecommerse Transaction

ciuffagianluca:
#### Types of changes

#### Checklist:


infantesimone (dismissed):
@ciuffagianluca well done! I have noticed an increase of pod startup time due to AI agent, for now I have increased the memory in dev but then we will do dedicated load tests.

infantesimone (approved):

PR #61, opened 2022-10-14
[CHK-402] Address structured logging

ciuffagianluca:
#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #70, opened 2022-10-27
Chk 402 - Chk 731 context thread solution

ciuffagianluca:
The purpose of this PR is to address the need for searchable, structured, and easy-to-read log objects in the Azure Monitor app insights. Instead of providing a template for the log message and search using string like operators, the attempt is to enrich the log itself using structured data to attach to the message.

This particular PR implements the "Context per MDC" solution. Like described in PR #61 the scope is to fill the MDC, but in this case, to fill it values are read from context filled in the mono chain of a transaction. There's an open point about how to read value from body and put it in the context, doing it in an interceptor method, not in every line we need to do it.
Leave open this PR until #61 is not conclude, as valid alternative to it.

---UPDATE---

Adopted this solution in place of PR #61 .
Read evaluation in [CHK-731](https://pagopa.atlassian.net/browse/CHK-731)

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/mdcutilities/MDCFilter.java
```diff
@@ -48,4 +68,40 @@ private Map<String, String> getTransactionId(ServerHttpRequest request) {
...
+            String rptId = signal.getContextView().get(RPT_ID);
+            String paymentContextCode = signal.getContextView().get(PAYMENT_CONTEXT_CODE);
+            String transactionId = signal.getContextView().get(TRANSACTION_ID);
+            try {
+                fillMDC(contextValue, rptId, paymentContextCode, transactionId);
+            } finally {
+                logStatement.accept(signal.get());
+            }
+        };
+    }
```
pietro-tota, line 98: remove this unused method

pietro-tota (commented):
Well done! Just some comment/observation

pietro-tota (approved):

pietro-tota (approved):

PR #78, opened 2022-11-11
[CHK-742] feat(error-handling): add error event to queue when failing to invoke Nodo's `closePaymentV2`

giovanniberti:
#### List of Changes

* Added error event to queue when failing to invoke Nodo's `closePaymentV2`
* Refactored transaction transitions tests

#### Motivation and Context

We need to handle retries in case of failure when contacting external systems. In this case, when invoking Nodo's `closePaymentV2` (i.e. finalizing the transaction), if a failure happens we enqueue a error event that will be handled by `pagopa-ecommerce-scheduler-service`.

In any case an event is saved into the event store, regardless of the outcome (closePayment successful with OK outcome, closePayment with KO outcome, error).

In case of error, an error event is enqueued to be processed at a later date, under the constraint that the current time is not *too near* the payment token expiration.

What "too near" means is controlled by three new configuration parameters, `TRANSACTIONS_RETRY_OFFSET`,  `CLOSURE_RETRY_INTERVAL` and `PAYMENT_TOKEN_VALIDITY_TIME`, as follows:
 * Payment token expiration is computed as `transactionCreationTime + PAYMENT_TOKEN_VALIDITY_TIME`
 * The expected retry time is `now() + CLOSURE_RETRY_INTERVAL`
 * The soft deadline for retrying is `expirationTime - TRANSACTION_RETRY_OFFSET`
 * If this expected retry time is after the expiration of the payment token (minus `TRANSACTION_RETRY_OFFSET`) we "clamp" the retry instant at the soft deadline
 * If we are already after the soft deadline then we don't enqueue the event at all and let the transaction expire

[Want a nice drawing? Check out this](https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/537e7d04d3cf1403e87f7a8bf4479ece9a8ac1e0/src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java#L131-L155)

- [x] Unit  tests
- [x] Local integration testing

#### Types of changes

#### Checklist:


giovanniberti:
A few design considerations/open questions:
 * The error event (`TransactionClosureErrorEvent`) covers **both** a redis failure or a Nodo failure, should we handle them differently/with different events?
 * I refactored the closure handler to use `@Autowired` parameters in the constructor to avoid using reflection in tests (looks nicer :)), but in other parts of the codebase we used reflection to inject `@Value`s. Should I stick to reflection or is this ok?

giovanniberti:
> * The error event (`TransactionClosureErrorEvent`) covers **both** a redis failure or a Nodo failure, should we handle them differently/with different events?

No

> * Consider the following scenario:
>       
>   * closure handler gets called, closePayment is invoked but the response is dropped -> an error event is saved to the event store and enqueued
>   * closure handler gets called again by the scheduler, closePayment is invoked and answers with 422 -> do we have any way to get the outcome from Nodo?
>   * Depending on the outcome, we should transition the transaction's state to either `CLOSED` or `CLOSURE_FAILED`
 
For now we just let the transaction expire instead

README.md
```diff
@@ -34,6 +34,9 @@ These are all environment variables needed by the application:
 | NODOPERPM_URI                                  | NodoPerPM connection URI                                                           | string |         |
 | NODOPERPM_READ_TIMEOUT                         | NodoPerPM read timeout                                                             | number |         |
 | NODOPERPM_CONNECTION_TIMEOUT                   | NodoPerPM connection timeout                                                       | number |         |
```
pietro-tota, line 36: Maybe there is missing the `TRANSACTION_CLOSURE_SENT_EVENT_QUEUE_NAME` system property?

pietro-tota (dismissed):
🚀 Well done! just some comment

README.md
giovanniberti: Good catch! Fixed in 43dc59d

pietro-tota (approved):

pietro-tota (approved):

PR #84, opened 2022-12-06
[CHK-826] - feat: set cardholder info into AuthorizationRequestData

pietro-tota:
#### List of Changes
Propagate card authorization informations to the TransactionRequestAuthorizationHandler

#### Motivation and Context
Card informations such as pan, cvv etc are necessary in order to perform card payment authorization

Junit test 

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/commands/data/AuthorizationRequestData.java
```diff
@@ -11,5 +13,9 @@ public record AuthorizationRequestData(
         String brokerName,
         String pspChannelCode,
         String paymentMethodName,
-        String pspBusinessName
+        String pspBusinessName,
```
ciuffagianluca, line 16: Could we improve this implementation using dtos inherited from RequestAuthorizationRequestDetailsDto interface

PR #99, opened 2022-12-19
[CHK-890] refactoring of transaction status info dto

ciuffagianluca:
#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/services/TransactionsService.java
```diff
@@ -137,10 +141,18 @@ public Mono<TransactionInfoDto> getTransactionInfo(String transactionId) {
...
                                                         .rptId(noticeCode.getRptId())
                                         ).toList()
                                 )
+                                .amountTotal(transaction.getAmountTotal())
+                                .feeTotal(transaction.getFeeTotal())
+                                .origin(
+                                        TransactionInfoDto.OriginEnum.valueOf(
+                                                Optional.ofNullable(transaction.getOrigin()).orElse(
+                                                        it.pagopa.ecommerce.commons.documents.Transaction.OriginType.UNKNOWN
+                                                ).toString()
```
giovanniberti, line 153: Isn't `valueOf` redundant?
```suggestion
Optional.ofNullable(transaction.getOrigin()).orElse(it.pagopa.ecommerce.commons.documents.Transaction.OriginType.UNKNOWN).toString()
```

giovanniberti: Moreover: I wouldn't even bother checking wether `origin` is null, given that we always populate this field when initializing a transaction its absence would be a bug, so in that case a `NullPointerException` would be fine

ciuffagianluca: No because the original value `it.pagopa.ecommerce.commons.documents.Transaction.OriginType` is from documents package of commons. Here I specified in the yaml file the same enum, but it comes from another package. 
So I had 2 chances:
1) take the same enum from commons package. This indeed could be fine, but I need help to include it in the yaml file
2) do it as it is now.

giovanniberti: You're right, my bad :)
1. I don't know if it's even possible, 2. is fine, but what do you think about this:

```suggestion
TransactionInfoDto.OriginEnum.valueOf(transaction.getOrigin()).toString()
```

The thing I'm worried about is by using `Optional` we would mask a potential bug

ciuffagianluca: I did it in this way 👍

api-spec/transactions-api.yaml
```diff
@@ -338,17 +335,28 @@ components:
...
+          $ref: "#/components/schemas/AmountEuroCents"
+        feeTotal:
+          $ref: "#/components/schemas/AmountEuroCents"
+        origin:
+          description: cart origin
+          enum:
+            - IO
+            - CHECKOUT
+            - CHECKOUT_CART
+            - UNKNOWN
```
infantesimone, line 354: why UNKNOWN?

api-spec/transactions-api.yaml
```diff
@@ -338,17 +335,28 @@ components:
...
           $ref: "#/components/schemas/TransactionStatus"
+        amountTotal:
+          $ref: "#/components/schemas/AmountEuroCents"
+        feeTotal:
+          $ref: "#/components/schemas/AmountEuroCents"
+        origin:
+          description: cart origin
+          enum:
+            - IO
+            - CHECKOUT
```
infantesimone, line 352: do you need to have `CHECKOUT` and `CHECKOUT_CART`?

src/main/java/it/pagopa/transactions/services/TransactionsService.java
```diff
@@ -137,10 +141,18 @@ public Mono<TransactionInfoDto> getTransactionInfo(String transactionId) {
...
-                                                        .authToken(null)
                                                         .rptId(noticeCode.getRptId())
                                         ).toList()
                                 )
+                                .amountTotal(transaction.getAmountTotal())
+                                .feeTotal(transaction.getFeeTotal())
+                                // FIXME in the next release of commons library we will delete the null checking
+                                .origin(
+                                        TransactionInfoDto.OriginEnum.valueOf(
+                                                transaction.getOrigin() == null ? "UNKNOWN"
```
infantesimone, line 152: for an invalid origin better to block the transaction, what do you think about it?

api-spec/transactions-api.yaml
ciuffagianluca: We have the same value in yaml file for document.
This value will be deleted, but, right now, we didn't know how to fill origin in transaction document, or better, we know, but we didn't yet implement it.

src/main/java/it/pagopa/transactions/services/TransactionsService.java
ciuffagianluca: We won't have invalid origin (e.g. null). In the next release of commons library it will be filled with UNKNOW when no values are available. I need to put here this check to make test success. With the next commons release this check will be deleted

api-spec/transactions-api.yaml
infantesimone: good point,  next refinement :-)

api-spec/transactions-api.yaml
ciuffagianluca: I found that values [here](https://pagopa.atlassian.net/wiki/spaces/I/pages/568099134/2022-10-26+-+Refinement+POST+transactions+gestione+carrello). Moreover they reflect the same values specified in commons package for transaction document.

infantesimone (approved):

giovanniberti (approved):

PR #101, opened 2022-12-20
[CHK-894] fix: nodo request building side-effect

pietro-tota:
#### List of Changes
Removed Nodo base requests build as DI beans and class attribute

#### Motivation and Context
This fix is needed since having base request as clients classes attribute cause side-effect when multiple concurrent requests override requests values with the latest processed ones

Local test elaborating 5 payment notice concurrent processing requests

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/configurations/NodoConfig.java
+++ b/src/main/java/it/pagopa/transactions/configurations/NodoConfig.java
@@ -1,42 +1,47 @@
 package it.pagopa.transactions.configurations;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import it.pagopa.generated.nodoperpsp.model.NodoAttivaRPT;
 import it.pagopa.generated.nodoperpsp.model.NodoVerificaRPT;
 import it.pagopa.generated.transactions.model.ActivatePaymentNoticeReq;
 import it.pagopa.generated.transactions.model.VerifyPaymentNoticeReq;
 import it.pagopa.transactions.utils.NodoConnectionString;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
 
 @Configuration
+@Slf4j
 public class NodoConfig {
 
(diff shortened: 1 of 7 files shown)


src/main/java/it/pagopa/transactions/configurations/NodoConfig.java
```diff
@@ -1,42 +1,47 @@
...
+    @Bean
+    public NodoConnectionString nodoConnectionString() {
+        NodoConnectionString nodoConnectionString = null;
+        try {
+            nodoConnectionString = new ObjectMapper()
+                    .readValue(nodoConnectionParamsAsString, NodoConnectionString.class);
+        } catch (JsonProcessingException e) {
+            // exception not logged here since it can expose sensitive information contained
+            // into nodo connection string
+            log.error("Exception parsing nodo connection string");
```
infantesimone, line 38: in this case we should not start the pod, and block the rolling update on the deployment on aks. in this way we block the deployment of the new app version, what do you think?

infantesimone: the question is, in case of invalid connection string, springboot application starts?

pietro-tota: Yes, you're absolutely right, fixed and tested with wrong connections string as parameter checking that application failed to start

src/main/java/it/pagopa/transactions/configurations/NodoConfig.java
```diff
@@ -1,42 +1,47 @@
...
+        }
+        return nodoConnectionString;
```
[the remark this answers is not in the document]
giovanniberti: (Closing as there is another conversation on this)

src/main/java/it/pagopa/transactions/configurations/NodoConfig.java
```diff
@@ -1,42 +1,47 @@
...
-                                                      ) String nodoConnectionParamsAsString,
-                                                      it.pagopa.generated.nodoperpsp.model.ObjectFactory objectFactoryNodoPerPsp
-    )
-            throws JsonProcessingException {
+    @Autowired
+    private it.pagopa.generated.transactions.model.ObjectFactory objectFactoryNodeForPsp;
 
-        NodoConnectionString nodoConnectionParams = nodoConnectionString(nodoConnectionParamsAsString);
+    @Bean
+    public NodoConnectionString nodoConnectionString() {
```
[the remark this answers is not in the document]
pietro-tota: NodoConnectionsString is marked as @Bean and it's treated as a Singleton by Spring framework

infantesimone (approved):

PR #102, opened 2022-12-20
Refactor logging to log transaction ids instead of RPT ids

giovanniberti:
#### Types of changes

#### Checklist:


src/test/java/it/pagopa/transactions/client/PaymentGatewayClientTest.java
```diff
@@ -324,7 +324,7 @@ void shouldThrowAlreadyProcessedOn401ForCreditCardWithXpay() throws JsonProcessi
         StepVerifier.create(client.requestXPayAuthorization(authorizationData))
                 .expectErrorMatches(
                         error -> error instanceof AlreadyProcessedException &&
-                                ((AlreadyProcessedException) error).getRptId()
+                                ((AlreadyProcessedException) error).getTransactionId()
                                         .equals(transaction.getNoticeCodes().get(0).rptId())
```
pietro-tota, line 328: Here should equality be tested against transactions id and not 1st notices RPT ID?

src/test/java/it/pagopa/transactions/client/PaymentGatewayClientTest.java
```diff
@@ -400,7 +400,7 @@ void shouldThrowAlreadyProcessedOn401ForPostePay() throws JsonProcessingExceptio
         StepVerifier.create(client.requestPostepayAuthorization(authorizationData))
                 .expectErrorMatches(
                         error -> error instanceof AlreadyProcessedException &&
-                                ((AlreadyProcessedException) error).getRptId()
+                                ((AlreadyProcessedException) error).getTransactionId()
                                         .equals(transaction.getNoticeCodes().get(0).rptId())
```
pietro-tota, line 404: Same as above

pietro-tota (approved):
Well done! Just some comment. Also there is a compilation error with the `TransactionAddUserReceiptHandler`, did you miss the genaral refactoring for this class?

src/test/java/it/pagopa/transactions/client/PaymentGatewayClientTest.java
giovanniberti: Whoops! Missed that 😅  Fixed in c84d279

src/test/java/it/pagopa/transactions/client/PaymentGatewayClientTest.java
giovanniberti: (you're right, see below, fixed in c84d279)

pietro-tota (approved):

PR #112, opened 2023-01-03
Chk 924 transaction origin header

pietro-tota:
#### List of Changes
Add POST transactions 'X-Client-Id' optional header evaluation to set transaction origin into transaction events.

#### Motivation and Context
This modification is need in order to set transaction origin value taking information from the request origin

Local test

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/api-spec/transactions-api.yaml
+++ b/api-spec/transactions-api.yaml
@@ -10,12 +10,10 @@ paths:
       parameters:
         - in: header
           name: X-Client-Id
-          schema:
-            type: string
-            enum: [IO,CHECKOUT,CHECKOUT_CART]
-            pattern: 'IO|CHECKOUT|CHECKOUT_CART'
           required: false
           description: Transaction origin (populated by APIM policy)
+          schema:
+            $ref: '#/components/schemas/ClientId'
       summary: Make a new transaction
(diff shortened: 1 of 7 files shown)


Fernando-Granato (approved):

api-spec/transactions-api.yaml
```diff
@@ -7,6 +7,14 @@ paths:
   /transactions:
     post:
       operationId: newTransaction
+      parameters:
+        - in: header
+          name: X-Client-Id
+          schema:
+            type: string
+            pattern: 'IO|CHECKOUT|CHECKOUT_CART'
```
giovanniberti, line 15: Bikeshedding: enum?

giovanniberti (approved):

api-spec/transactions-api.yaml
pietro-tota: I changed the type to enumeration but the pattern is still needed since header is generated as string object and removing pattern descriptor will cause generated code to not have any kind of content validation

giovanniberti (approved):

giovanniberti (approved):

PR #114, opened 2023-01-05
fix: refactoring origin to client id + commons update

pietro-tota:
#### List of Changes

1. Update commons library version with the latest one (refactored document origin to clientId)
2. Refactored origin field to clientId into openapi specifications

#### Motivation and Context
Those modifications are needed in order to refactor `origin` to a meaningful `clientId` name

#### How Has This Been Tested?
Local test (using ecommerce local and performing a new transactions checking stored events and returned responses)

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #115, opened 2023-01-05
[CHK-933] fix(node): fix Redis population with null values in case of cache miss

giovanniberti:
#### List of Changes

 * fix Redis population with null values in case of cache miss

#### Motivation and Context

This PR fixes an issue occurring when calling `POST /transactions` directly or when encountering a cache miss when invoking this endpoint. 
Previously, the payment information data saved on Redis was not populated from Nodo's activation response. 
Upon a cache miss `null` values would be saved to Redis and could wreak havoc on later payment flows for the same notice (as the Redis cache would be incorrectly populated).

#### Types of changes

#### Checklist:

Depends on: pagopa/pagopa-ecommerce-commons#18
Depends on: pagopa/pagopa-ecommerce-transactions-service#114

### DIFF
--- a/src/main/java/it/pagopa/transactions/utils/NodoOperations.java
+++ b/src/main/java/it/pagopa/transactions/utils/NodoOperations.java
@@ -3,10 +3,7 @@ package it.pagopa.transactions.utils;
 import it.pagopa.ecommerce.commons.domain.IdempotencyKey;
 import it.pagopa.ecommerce.commons.domain.RptId;
 import it.pagopa.ecommerce.commons.repositories.PaymentRequestInfo;
-import it.pagopa.generated.nodoperpsp.model.EsitoNodoAttivaRPTRisposta;
-import it.pagopa.generated.nodoperpsp.model.NodoAttivaRPT;
-import it.pagopa.generated.nodoperpsp.model.NodoTipoCodiceIdRPT;
-import it.pagopa.generated.nodoperpsp.model.NodoTipoDatiPagamentoPSP;
+import it.pagopa.generated.nodoperpsp.model.*;
 import it.pagopa.generated.transactions.model.ActivatePaymentNoticeReq;
 import it.pagopa.generated.transactions.model.CtQrCode;
 import it.pagopa.generated.transactions.model.StOutcome;
@@ -22,6 +19,7 @@ import reactor.core.publisher.Mono;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.security.SecureRandom;
+import java.time.LocalDate;
 import java.util.Optional;
 
 @Slf4j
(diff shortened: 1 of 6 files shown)


src/main/java/it/pagopa/transactions/utils/NodoOperations.java
```diff
@@ -120,6 +104,9 @@ private Mono<String> nodoActivationForNM3PaymentRequest(
         request.setAmount(amount);
         request.setQrCode(qrCode);
         request.setIdempotencyKey(idempotencyKey);
+
+        LocalDate dueDate = LocalDate.now().plusYears(100);
```
giovanniberti, line 108: Open point: right now `dueDate` is modeled incorrectly inside `PaymentRequestInfo` (there is a wrong `@Pattern` annotation and most importantly a `@Nonnull` one). Here i populate it with a dummy value but `null` should be the correct choice. 

Should we:
 - have a PR remove the annotations on `dueDate` and then use `null` here
 - use `null` here directly (personally I don't like it very much as it would create an inconsistency between the model and how it's actuallty used)
 - keep the dummy value
?

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
```diff
@@ -181,8 +176,31 @@ private Mono<String> nodoActivationForUnknownPaymentRequest(
...
+                                    .orElse("");
+
                             return StOutcome.OK.value().equals(nodoAttivaRPTRResponse.getEsito())
-                                    ? Mono.just("")
+                                    ? Mono.just(
+                                            new PaymentRequestInfo(
+                                                    rptId,
+                                                    rptId.getFiscalCode(),
+                                                    paName,
+                                                    description,
```
giovanniberti, line 195: Same thing here, should we make `paName` and `description` optional?

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
pietro-tota: I would say that the first option is the better one.

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
pietro-tota: I would make them optional since they are optional in received nodo response.

pietro-tota (approved):
Well done!🚀

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
giovanniberti: Option 1: pagopa/pagopa-ecommerce-commons#18

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
giovanniberti: See: pagopa/pagopa-ecommerce-commons#18

pietro-tota (approved):

pietro-tota (approved):

PR #116, opened 2023-01-12
fix: transaction amount

pietro-tota:
#### List of Changes
Fix transaction amount: conversion between eurocent amount and it's BigDecimal representation is wrong and cents part is lost during conversion

#### Motivation and Context
This fix is needed in order to take request amount correctly

local test

#### Types of changes

#### Checklist:


giovanniberti (approved):
👀

infantesimone:
Reminder: when we do a fix like in this case, it's better to add a unit test anyway :-)

pietro-tota:
> Reminder: for a fix like this, it's better to add a unit test anyway :-)

Yes, you are right, I'll add it in a separate PR

PR #117, opened 2023-01-12
[CHK-976] fix: return code in case of auth request amount mismatch

pietro-tota:
#### List of Changes
Map transaction auth request amount mismatch error to 409 Conflict

#### Motivation and Context
Transaction auth request with mismatching amount (the request and the transaction stored one) are mapped to 404 not found error instead of 409 conflict, returning a misleading error code and message to the caller.

Local test and junit test

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/services/TransactionsService.java
```diff
@@ -175,11 +176,15 @@ public Mono<RequestAuthorizationResponseDto> requestTransactionAuthorization(
                                     transactionId
                             );
                             return !amountTotal.equals(requestAuthorizationRequestDto.getAmount())
-                                    ? Mono.empty()
+                                    ? Mono.error(
+                                            new TransactionAmountMismatchException(
+                                                    requestAuthorizationRequestDto.getAmount(),
+                                                    amountTotal
+                                            )
+                                    )
```
pietro-tota, line 184: Should we treat this error case as a transactions not found exception or a bad request?

giovanniberti: If we are to differentiate the two cases I'd vote for 422 and reserve 400 for purely formal errors in the request body

MarcoAGreco: bikeshedding time come back :)  
I'd vote for a generic 400 bad request, just to avoid to give more then the required information to the frontend

MarcoAGreco (approved):

src/main/java/it/pagopa/transactions/services/TransactionsService.java
pietro-tota: Restored to generic bad request return code, shall we go?

giovanniberti: Voted for 409

infantesimone: sounds good

src/main/java/it/pagopa/transactions/exceptions/TransactionAmountMismatchException.java
```diff
@@ -0,0 +1,15 @@
+package it.pagopa.transactions.exceptions;
+
+public class TransactionAmountMismatchException extends RuntimeException {
+
+    public TransactionAmountMismatchException(
+            Integer requestAmount,
+            Integer transactionAmount
+    ) {
+        super(
+                "Transaction amount mismatch! Request amount: [%s], Transaction amount: [%s]"
```
infantesimone, line 10: focus point, is it correct to return this message to the client? it is important to log (with a dedicated exception) but the client doesn't have to know the correct amount in the storage

infantesimone (changes requested):

src/main/java/it/pagopa/transactions/exceptions/TransactionAmountMismatchException.java
pietro-tota: You're right. Modified into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/117/commits/8c15e62e172a81ad3ac5e21333338166bae130a6

infantesimone (approved):

PR #120, opened 2023-01-20
[CHK-958] feat: jwt token generation

pietro-tota:
Depends on pagopa/pagopa-ecommerce-local#34
#### List of Changes
Add JWT token generation on new transaction requests

#### Motivation and Context
Removed Session service call to generate token and moved into transaction service module.
This modification permits to remove completely Session Service module invocation on new transaction.
Token generation is now demanded to transaction service that will return JWT token on each new transaction.

Token validity check is demanded to APIM policy that will verify the token validity for each subsequence calls to transaction for that transaction id.

Generated token contains the following claims:
```
{
  "transactionId": "e0501f8f-423f-4dbc-a00b-0749e1068330",
  "jti": "6ffa2ea2-f165-41c4-b9e2-93662e1f19ea",
  "iat": 1674555379,
  "exp": 1674557179
}
```
Having custom transactionId claim into the generated token allows to check any received token against the request transaction id to check for equality.
iat and exp claims have been added in respect of the sessions service generated token so that check against token duration validity can be performed.

Token signing key and token validity time have been parametrized with the following system properties:

- JWT_SECRET --> the secret key to be used for token signature
- JWT_TOKEN_VALIDITY_TIME_MILLIS --> the token validity expressed in millisecond

Session service integration will be removed in PR #121.

Local test + junit test

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/commands/TransactionActivateCommand.java
```diff
@@ -10,15 +10,20 @@
...
+    private final Transaction.ClientId clientId;
+
+    private final String transactionId;
 
     public TransactionActivateCommand(
             RptId rptId,
             NewTransactionRequestDto data,
-            Transaction.ClientId clientId
+            Transaction.ClientId clientId,
+            String transactionId
```
giovanniberti, line 21:
```suggestion
            TransactionId transactionId
```

src/main/java/it/pagopa/transactions/exceptions/JWTTokenGenerationException.java
```diff
@@ -0,0 +1,8 @@
+package it.pagopa.transactions.exceptions;
+
+public class JWTTokenGenerationException extends Exception {
```
giovanniberti, line 3: If this is not meant to be used as a checked exception we should extend `RuntimeException`

src/main/java/it/pagopa/transactions/utils/JwtTokenUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+    }
+
+    private String generateNonce() {
+        StringBuilder nonce = new StringBuilder();
+        for (int curr = 0; curr < 15; curr++) {
+            nonce.append(secureRandom.nextInt(10));
+        }
+        return nonce.toString();
+    }
```
giovanniberti, line 60: Just a quick comment (that maybe should be included in code to validate our assumptions in the future 🙂)

There are $10^{15}$ possibilities for a nonce
Each nonce is valid for 30 minutes (a session duration, currently the same as the duration of a payment token)

Assuming 100 HTTP req/s for an attacker (for a single IP address it's an infrastructure constraint we put, in a more advanced threat model we could assume a distributed attack), we have
100 rps * (30 * 60) = 180 k guesses

$p(\text{1 right guess}) = \frac{1}{10^{15}}$
$p(\text{right guess before expiry}) = \frac{1.8 * 10^5}{10^{15}} = 1.8 * 10^{-10} \approx 2^{-29}$

We'll have to discuss with security whether this value is enough, but given how short these tokens' durations are as a first approximation I'd say it's ok

giovanniberti (changes requested):

src/main/java/it/pagopa/transactions/utils/JwtTokenUtils.java
pietro-tota: Generated tokens are signed with HMAC key retrieved from KV that must be at least 256-bit long as stated by `Keys.hmacShaKeyFor` that will throw `WeakKeyException` for shorter keys. 
As for now, if the secret configured key is an invalid base64 key or shorter than the length expressed above then the module startup fails.

Considering that the tokens are signed with HMAC and that into the token body is present also the transactionId claim that is a random UUID generated for each transaction I would say that is not a problem.

Beside that, since [jti token claim](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.7) is intended to be an unique id for the token (mainly used for token revocations) I changed the jti claim to be an UUID so that it can be identified univocally.

src/main/java/it/pagopa/transactions/commands/TransactionActivateCommand.java
pietro-tota: Fixed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/120/commits/b3550051b8e712d78f388ea21c8035cecf04fc14

src/main/java/it/pagopa/transactions/exceptions/JWTTokenGenerationException.java
pietro-tota: Fixed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/120/commits/b3550051b8e712d78f388ea21c8035cecf04fc14

infantesimone:
pay attention to integration tests

infantesimone (approved):

PR #121, opened 2023-01-20
[CHK-990] feat: remove session service integration

pietro-tota:
Depends on pagopa/pagopa-ecommerce-transactions-service#120
#### List of Changes
Removed session service invocation

#### Motivation and Context
Session service invocation is no more needed since session token is now generated by transactions-service

Local test

#### Types of changes

#### Checklist:


infantesimone:
pay attention to integration tests

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -58,24 +53,22 @@ public TransactionActivateHandler(
...
         this.paymentTokenTimeout = paymentTokenTimeout;
         this.transactionActivatedQueueAsyncClient = transactionActivatedQueueAsyncClient;
     }
 
-    public Mono<Tuple3<Mono<TransactionActivatedEvent>, Mono<TransactionActivationRequestedEvent>, SessionDataDto>> handle(
-                                                                                                                           TransactionActivateCommand command
+    public Mono<Tuple2<Mono<TransactionActivatedEvent>, Mono<TransactionActivationRequestedEvent>>> handle(
+                                                                                                           TransactionActivateCommand command
     ) {
-        final String transactionId = UUID.randomUUID().toString();
```
infantesimone, line 78: the `TransactionActivateCommand` doesn't have the `transactionId` which is generated here, why delete it?

src/main/java/it/pagopa/transactions/configurations/WebClientsConfig.java
```diff
@@ -257,7 +227,7 @@ public it.pagopa.generated.notifications.v1.api.DefaultApi notificationsServiceW
                         )
                 );
 
-        WebClient webClient = ApiClient.buildWebClientBuilder().clientConnector(
+        WebClient webClient = it.pagopa.generated.notifications.v1.ApiClient.buildWebClientBuilder().clientConnector(
```
infantesimone, line 230: Why this refactoring?

src/main/java/it/pagopa/transactions/controllers/TransactionsController.java
```diff
@@ -53,7 +57,11 @@ public Mono<ResponseEntity<NewTransactionResponseDto>> newTransaction(
                                                                           ClientIdDto clientIdDto,
                                                                           ServerWebExchange exchange
     ) {
-        return newTransactionRequest
+        TransactionId transactionId = new TransactionId(UUID.randomUUID());
+        String jwtToken = jwtTokenUtils.generateToken(transactionId);
```
infantesimone, line 61: In this way a jwt token is generated even for requests with invalid notice codes, maybe it's better to generate the jwt token after pagoPA nodo activation? where was session-service client invocation?

infantesimone (changes requested):

src/main/java/it/pagopa/transactions/controllers/TransactionsController.java
pietro-tota: Yes, you're right, I'll move the token generation in place of the session service invocation

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
pietro-tota: The transaction id is now generated into the TransactionController just because it's needed for JWT token generation. I'm going to move it back here since token will be generated in place of session service call

src/main/java/it/pagopa/transactions/controllers/TransactionsController.java
pietro-tota: Modification addressed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/121/commits/f1619dca36997c33067c1f7b842e05e64942c277

I'he done this modifications only in this PR that take in charge the sessions-service remove and not in #120 that have the focus on JWT token generation logic porting

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
pietro-tota: Resolved into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/121/commits/f1619dca36997c33067c1f7b842e05e64942c277

infantesimone (approved):

PR #127, opened 2023-01-31
[CHK-1032] update auth-reqeust and add check for threedsdata in transactionService

ciuffagianluca:
update auth-reqeust adding threedsdata parameter.

Add check for threedsdata in transactionService

**Brand field in cardAuthRequest details is temporary and will be deleted when the brand will be provided inside pgs.**

#### List of Changes

- auth-request post body definition
- test

#### Motivation and Context

In order to address integration with vpos client the auth-request post need more data like 3dsdata and card brand. For this field a temporary solution has been adopted and the brand is passed from front-end. We'll remove this field from cardDetails when the pgs service will be able to evaluate the brand itself from the card pan.

#### Types of changes

#### Checklist:


api-spec/transactions-api.yaml
```diff
@@ -392,6 +392,9 @@ components:
             mapping:
               postepay: "#/components/schemas/PostePayAuthRequestDetails"
               card: "#/components/schemas/CardAuthRequestDetails"
+        threeDsData:
```
infantesimone, line 395: It should be a `CardAuthRequestDetails` field

api-spec/transactions-api.yaml
```diff
@@ -436,18 +439,23 @@ components:
         holderName:
           type: string
           description: The card holder name
+        brand:
```
infantesimone, line 442: This field is already modeled in `CardAuthRequestDetails`

infantesimone (changes requested):
Open point, 3dsdata mandatory only in the case of `pgsId` equal to vpos

api-spec/transactions-api.yaml
ciuffagianluca: I didn't find it in the dto definition, so that I added it here in the definition of the dto itself.

api-spec/transactions-api.yaml
ciuffagianluca: Are you sure about this point? From my point of view 3dsdata field is not a property of the _creditcard_ object. Thinking when we will manage a wallet, I don't think that the 3dsdata will be part of the credit card, while it will be for sure part of the payment authorization step.

ciuffagianluca:
> Open point, 3dsdata mandatory only in the case of `pgsId` equal to vpos

I modeled this conditional mandatory in transaction service. But I'm not sure it is a good way to do it.

infantesimone (approved):

PR #138, opened 2023-02-14
feat: CHK-1078 add idempKey creation before activatePaymentNotice

infantesimone:
#### List of Changes

- `idempotencyKey` for `activatePaymentNotice` given fiscal code of ecommerce platform, so `IdBrokerPSP`;
- `idempotencyKey` creation before `activatePaymentNotice` and saved in redis cache .

#### Motivation and Context

The goal of this PR is to generate the `idempotencyKey` before `activatePaymentNotice`, and save it contextually in the redis cache with the rptId key, in order to use in retries.
depends on https://github.com/pagopa/pagopa-ecommerce-commons/pull/37

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
@@ -115,10 +115,7 @@ public class TransactionActivateHandler
                                                                             requestInfo -> isValidIdempotencyKey(
                                                                                     requestInfo.idempotencyKey()
                                                                                             .rawValue()
-                                                                            )
-                                                                                    && isValidPaymentToken(
-                                                                                            requestInfo.paymentToken()
-                                                                                    )
+                                                                            ) && isValidPaymentToken(requestInfo.paymentToken())
                                                                     )
                                                                     .orElseGet(
                                                                             () -> {
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
@@ -127,7 +127,6 @@ public class TransactionActivateHandler
                                                                                 null,
                                                                                 null,
                                                                                 null,
-                                                                                true,
                                                                                 null,
                                                                                 new IdempotencyKey(
                                                                                         nodoOperations
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
@@ -56,8 +56,6 @@ public class TransactionActivateHandler
     @Value("${nodo.parallelRequests}")
     private int nodoParallelRequests;
 
-    private static final String PSP_PAGOPA_ECOMMERCE_FISCAL_CODE = "00000000000";
-
     @Autowired
     public TransactionActivateHandler(
             PaymentRequestsInfoRepository paymentRequestsInfoRepository,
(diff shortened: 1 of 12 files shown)


src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -94,6 +95,53 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
...
                                 )
+                        ).flatMap(
+                                paymentRequest -> {
+                                    final PaymentNoticeInfoDto paymentNotice = paymentRequest.getT1();
+                                    final Optional<PaymentRequestInfo> maybePaymentRequestInfo = paymentRequest
+                                            .getT2();
+                                    return Mono.just(
+                                            Tuples.of(
+                                                    paymentNotice,
+                                                    Optional.of(
```
pietro-tota, line 106: Since paymentRequestInfo is always initialized even in case of cache miss can we remove the Optional from the returned Tuple here and return directly the payment request info obtained from the Optional.orElseGet chain?

pietro-tota (approved):
🚀 just a comment

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -94,6 +95,53 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
...
+                                                                                        null,
+                                                                                        true,
+                                                                                        null,
+                                                                                        new IdempotencyKey(
+                                                                                                nodoOperations
+                                                                                                        .getEcommerceFiscalCode(),
+                                                                                                nodoOperations
+                                                                                                        .generateRandomStringToIdempotencyKey()
+                                                                                        )
+                                                                                );
```
giovanniberti, line 133: If we decide that these fields can be `null` I would add a `@Nullable` annotation on them inside `PaymentRequestInfo` (and maybe a constructor with non-null fields only?)

giovanniberti (changes requested):

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
infantesimone: correct! for a payment request we can also save only `<rptId,idempotencyKey>`, in order to handle retry with the same idempotencyKey. So, I have to to update `ecommerce-commons` -> https://github.com/pagopa/pagopa-ecommerce-commons/blob/main/src/main/java/it/pagopa/ecommerce/commons/repositories/PaymentRequestInfo.java

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
infantesimone: correct :-)

infantesimone:
depends on https://github.com/pagopa/pagopa-ecommerce-commons/pull/37

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
giovanniberti: Solved with pagopa/pagopa-ecommerce-commons#37

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
giovanniberti: Solved with d6101c8

pietro-tota (approved):

pietro-tota:
Watch out broken integration tests

PR #140, opened 2023-02-15
feat: CHK-1112 uuid in bytes base64 encoding

infantesimone:
#### List of Changes

- uuid in bytes base64 encoding
- only for `PATCH transactions/:idTransaction/auth-requests` have idTransaction as bytes base64 in input

#### Motivation and Context

The goal of this PR is to send to [PGS](https://github.com/pagopa/pagopa-payment-transactions-gateway) transactionId as bytes base64, in order to have a transactionId representation with  less 30 chars according to [XPAY specs](https://ecommerce.nexi.it/specifiche-tecniche/codicebase.html).
The reason why it's a good strategy is described in the comments 😁 (thanks).

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/utils/UUIDUtils.java
```diff
@@ -0,0 +1,24 @@
...
+public class UUIDUtils {
+
+    public String uuidToBase64(UUID uuid) {
+        ByteBuffer bb = ByteBuffer.wrap(new byte[16]);
+        bb.putLong(uuid.getMostSignificantBits());
+        bb.putLong(uuid.getLeastSignificantBits());
+        return Base64.encodeBase64URLSafeString(bb.array());
+    }
+
+    public UUID uuidFromBase64(String str) {
```
pietro-tota, line 19: For perform this conversion javascripts side we can use a similar approach used [here](https://stackoverflow.com/questions/47505620/javas-uuid-nameuuidfrombytes-to-written-in-javascript)

I just made a playable version to test the conversion from transactions service generated base64 uuid back to UUID string format [here](https://playcode.io/1195321) and it seems to work correctly (the last comment in code is the output is taken making a main function using your utility)

ciuffagianluca: We need to focus these 3 points to be sure that this conversion works right:
1. Base64 encondig function to be a bijection: this is assured by math principles of Base64 encoding. Starting from a codomain's value Y we are sure that we can found an inverted function (and we know it, the decoding) that get us the domain's value X we encode to obtain Y.
2. bytearray function to be a bijection: since it is only a different math representation of the same bits, we are sure that from 2 long values representing most significant and least significant bits of a UUID we can obtain the original UUID from which we split to obtain them.
3. implement an algorithm to decode the encoded UUID: the solution showed by @pietro-tota seems to work fine. Again it is only a different math representation, so math help us.

MarcoAGreco: Base64 is not a bijection (this is a consequence of the padding), but this is not a problem if we use the decoded value for processing. The other two points are satisfied so I think that we can use this approach.

```
const b64_encoded_str = "AAA="; // Zero byte
const b64_ecnoded_str_2 = "AAB="; // Zero byte

console.log(`str1: ${b64_encoded_str}\tstr2: ${b64_ecnoded_str_2}`)
console.log(`str1 === str2: ${b64_encoded_str === b64_ecnoded_str_2}`);

const decode_str = atob(b64_encoded_str);
const decode_str2 = atob(b64_ecnoded_str_2);

// Only the 12-msb of the string are used during the encoding process because of the padding
// that cause a possible collision during the decoding process
console.log(`decoded(str1) === decoded(str2): ${decode_str === decode_str2}`);
```
 
Try it [here](https://jsfiddle.net/Lwufadqo/23/)

pietro-tota (dismissed):
🚀

PR #143, opened 2023-02-17
[CHK-1123] add decoder for 3ds data

ciuffagianluca:
#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
+++ b/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
@@ -45,7 +45,6 @@ public class PaymentGatewayClient {
     public Mono<PostePayAuthResponseEntityDto> requestPostepayAuthorization(
                                                                             AuthorizationRequestData authorizationData
     ) {
-
         return Mono.just(authorizationData)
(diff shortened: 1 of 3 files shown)


infantesimone (approved):

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -170,7 +169,7 @@ public Mono<VposAuthResponseDto> requestCreditCardAuthorization(AuthorizationReq
                                         .holder(cardData.getHolderName())
                                         .securitycode(cardData.getCvv())
                                         .isFirstPayment(true) // TODO TO BE CHECKED
-                                        .threeDsData(cardData.getThreeDsData())
+                                        .threeDsData(new String(Base64.getDecoder().decode(cardData.getThreeDsData())))
```
pietro-tota, line 172: Use here the String constructor with explicit encoding, the used one here use the system default one. 
Better to use here the explicit expected encoding such as UTF-8 to avoid problems with special characters encoding

pietro-tota (changes requested):

pietro-tota:
Add a brief PR description

ciuffagianluca:
Close, no more need to decode data

PR #144, opened 2023-02-20
Fix removed unused variables

Fernando-Granato:
#### List of Changes
Removed unused variables and changed queue names in configuration files.

#### Types of changes

#### Checklist:


src/main/resources/application.properties
```diff
@@ -36,8 +36,6 @@ management.endpoint.health.show-details=always
 management.health.circuitbreakers.enabled=true
 
 azurestorage.connectionstring=${ECOMMERCE_STORAGE_CONNECTION_STRING}
-azurestorage.queues.transactionauthrequestedtevents.name=${TRANSACTION_AUTHORIZATION_REQUESTED_EVENT_QUEUE_NAME}
```
pietro-tota, line 39: Are you sure you have to remove this configuration property?
It's used here https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/b359c976d23e58a7ce393cd19daee6ae6de93328/src/main/java/it/pagopa/transactions/configurations/AzureStorageConfig.java#L14 and the builded queueClient is used here https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/b359c976d23e58a7ce393cd19daee6ae6de93328/src/main/java/it/pagopa/transactions/commands/handlers/TransactionRequestAuthorizationHandler.java#L37

PR #148, opened 2023-02-21
feat: add TransactionClosedEvent send on queue

pietro-tota:
#### List of Changes

Add logic for send TransactionClosedEvent on the closure event queue in case the transactions was previously authorized and Nodo closePayment response outcome is KO or the HTTP response status code is an unrecoverable ones (such as 400).
This event is sent with visibility timeout set to 0 so that events dispatcher can process it as soon as possible and perform transaction rollback logic.

#### Motivation and Context

This newly written event will be read by ecommerce event-dispatcher in order to perform transaction refund.

Local junit tests + local test using ecommerce local mocking PGS and Nodo responses

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
@@ -146,10 +146,16 @@ public class TransactionSendClosureHandler extends
                     log.info("Invoking closePaymentV2 for RptId: {}", paymentNotice.rptId().value());
                     return nodeForPspClient.closePaymentV2(closePaymentRequest)
                             .flatMap(
-                                    response -> processClosureEvent(
+                                    response -> buildAndSaveClosureEvent(
                                             command,
                                             transactionAuthorizationCompletedData.getAuthorizationResultDto(),
                                             response.getOutcome()
+                                    )
+                            )
+                            .flatMap(
+                                    event -> sendClosureEvent(
+                                            event,
+                                            transactionAuthorizationCompletedData.getAuthorizationResultDto()
                                     )
                             )
                             .map(Either::<TransactionClosureErrorEvent, TransactionEvent<TransactionClosureData>>right)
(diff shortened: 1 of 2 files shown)


src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
```diff
@@ -270,15 +267,15 @@ private ClosePaymentRequestV2Dto.OutcomeEnum authorizationResultToOutcomeV2(
         }
     }
 
-    private Mono<TransactionEvent<TransactionClosureData>> buildClosureEvent(
+    private Mono<TransactionEvent<TransactionClosureData>> processClosureEvent(
```
giovanniberti, line 270: This method does the following things:
 * create a new closure event, either closed or closureFailed
 * decide whether it's enqueueable
 * enqueue it

I would split the method (especially the first two points from the rest), as right now it's not very clear from the outside what it's doing

src/test/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandlerTest.java
```diff
@@ -1647,6 +1684,598 @@ void shouldEnqueueErrorEventOnNodoRecoverableFailure() {
...
+        TransactionActivatedEvent transactionActivatedEvent = new TransactionActivatedEvent(
+                transactionId.value().toString(),
+                new TransactionActivatedData(
+                        email.value(),
+                        PaymentNotices,
+                        faultCode,
+                        faultCodeString,
+                        Transaction.ClientId.CHECKOUT
+                )
+        );
```
giovanniberti, line 1715: Can you use utilities from `TransactionTestUtils`?

giovanniberti (changes requested):

src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
pietro-tota: Fixed in https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/148/commits/e8807b65afebce1c60687eab81d4e543b0efadf7

src/test/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandlerTest.java
pietro-tota: Fixed in https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/148/commits/e8807b65afebce1c60687eab81d4e543b0efadf7

giovanniberti (approved):

PR #150, opened 2023-02-22
Chk 1140 use confidential data for email field

pietro-tota:
Depends on the following PRs:

- infra pagopa/pagopa-infra#747 for secret creation in ecommerce KV
- commons pagopa/pagopa-ecommerce-commons#35 for email field client-side encryption

#### List of Changes

Update ecommerce-commons to the branch in which emails is protected with Confidential<Email> field.
Created utility logics to handle Confidential<Email> field for encryption (activation phase, where user email is received as plain-text into POST transaction request and is saved as encrypted value into eventstore) and decryption in every point where clear email is needed such as PGS VPOS and notification-service  requests building. 

#### Motivation and Context

With those modifications user email, which is a sensitive data, is no more store as plain-text on mongo eventstore collection but encrypted and only modules which have access to secret AES encryption key can decrypt email back to the original plain-text

Junit tests and local tests. Local tests were performed mocking PGS and notifications-service and checking that email is stored encrypted into eventstore TRANSACTION_ACTIVATED event and that email is decrypted successfully for PGS and notification-service requests set mail into those requests with the original value which is passed during activation process.
 

#### Types of changes

#### Checklist:


pom.xml
```diff
@@ -453,8 +453,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>confidential-email</scmVersion>
```
pietro-tota, line 457: To be rollback after ecommerce-commons PR merge

.env.example
```diff
@@ -58,4 +58,7 @@ TRANSACTIONS_RETRY_OFFSET=60
 CLOSURE_RETRY_INTERVAL=30
 
 JWT_SECRET=ODMzNUZBNTZENDg3NTYyREUyNDhGNDdCRUZDNzI3NDMzMzQwNTFEREZGQ0MyQzA5Mjc1RjY2NTQ1NDk5MDMxNzU5NDc0NUVFMTdDMDhGNzk4Q0Q3RENFMEJBODE1NURDREExNEY2Mzk4QzFEMTU0NTExNjUyMEExMzMwMTdDMDk
-JWT_TOKEN_VALIDITY_TIME_MILLIS=600000
\ No newline at end of file
+JWT_TOKEN_VALIDITY_TIME_MILLIS=600000
+
+CONFIDENTIAL_DATA_MANAGER_KEY=EWsfdHUIDwLwYwC0zIMlrw==
```
giovanniberti, line 63: I don't like the fact that this env variable is referring to an implementation detail, what about `ENCRYPTION_SECRET_KEY`? Maybe it's a bit too generic...

.env.example
```diff
@@ -58,4 +58,7 @@ TRANSACTIONS_RETRY_OFFSET=60
...
+CONFIDENTIAL_DATA_MANAGER_ALGORITHM=AES
```
giovanniberti, line 64: This should **always** be AES, as currently it's the only cipher `ConfidentialDataManager` understands. Maybe it warrants a change of interface in `ecommerce-commons`, what do you think?

src/main/java/it/pagopa/transactions/utils/MailConfidentialDataUtility.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+
+@Component
+@Slf4j
+public class MailConfidentialDataUtility {
```
giovanniberti, line 21: For consistency

```suggestion
public class MailConfidentialDataUtils {
```

But personally I like

```suggestion
public class ConfidentialDataMailUtils {
```

or 

```suggestion
public class ConfidentialMailUtils {
```

even more :)

src/test/java/it/pagopa/transactions/projections/handlers/ClosureErrorProjectionHandlerTest.java
```diff
@@ -58,9 +56,46 @@ void shouldHandleProjection() {
...
+     * -6, 121, -55, 83, -73, 68, 56, 32, -81, -109, -82, -71],
+     * iv=javax.crypto.spec.IvParameterSpec@5568c66f},
+     * opaqueData=6z0M1JlS6m2XZKU2ryfSiwkPStx+e30eeH/0l/8vLeEhhBXch0JQ4i0kYiN0asY=],
+     * status=CLOSURE_ERROR, feeTotal=null,
+     * creationDate=2023-02-22T10:28:02.595734+01:00[Europe/Rome],
+     * paymentNotices=[PaymentNotice(paymentToken=paymentToken,
+     * rptId=77777777777111111111111111111, description=description, amount=100,
+     * paymentContextCode=paymentContextCode)]))
+     */
+
```
giovanniberti, line 94: 🫣

.env.example
pietro-tota: Yes, if we are not planning to change encryption algorithm I think that this parameter can be removed altogether

src/test/java/it/pagopa/transactions/projections/handlers/ClosureErrorProjectionHandlerTest.java
pietro-tota: Ahahahahaha, whoops, some residual manual check performed on unit tests 🤣

.env.example
pietro-tota: I don't know, maybe `EMAIL_ENCRYPTION_KEY` is a better choice just to empathize that this encryption key is used only for mail encryption and that, for future confidential data a new key must be used rather that this. What do you think?

pietro-tota: Fixed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/150/commits/c0ff765bde5bc8d25a69ff0d7e92d77dae01ef6f

.env.example
pietro-tota: Fixed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/150/commits/c0ff765bde5bc8d25a69ff0d7e92d77dae01ef6f

src/main/java/it/pagopa/transactions/utils/MailConfidentialDataUtility.java
pietro-tota: Fixed into https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/150/commits/c0ff765bde5bc8d25a69ff0d7e92d77dae01ef6f

giovanniberti (approved):

PR #157, opened 2023-02-28
[CHK-1129] Mask cvv pan email log

ciuffagianluca:
Mask data about cvv pan and email when they could be prionted in log message

#### List of Changes

Added logback configuration to handle custom layout message

#### Motivation and Context

we don't want to log sensitive data

Tested in local in environment
Added junit test

#### Types of changes

#### Checklist:


src/test/java/it/pagopa/transactions/utils/LogMaskTests.java
```diff
@@ -0,0 +1,128 @@
...
+                                new TransactionAmount(100),
+                                new TransactionDescription("description"),
+                                new PaymentContextCode(null)
+                        )
+                ),
+                null,
+                null,
+                null,
+                it.pagopa.ecommerce.commons.documents.v1.Transaction.ClientId.CHECKOUT
+        );
```
giovanniberti, line 88: Use `TransactionTestUtils.transactionActivated()` 🙂

ciuffagianluca: Funny!

src/main/java/it/pagopa/transactions/utils/LogMaskerPatternLayout.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.stream.IntStream;
+
+public class LogMaskerPatternLayout extends PatternLayout {
+    private Pattern patternString;
+    private List<String> maskPatterns = new ArrayList<>();
+
+    public void addMaskPattern(String maskPattern) {
+        maskPatterns.add(maskPattern);
+        patternString = Pattern.compile(maskPatterns.stream().collect(Collectors.joining("|")), Pattern.MULTILINE);
+    }
```
pietro-tota, line 21: Where will be this method be called at runtime to add the mask patterns?

ciuffagianluca: It is not directly called by any other point in the project, but it is necessary to make the logback configuration file working, when it add the mask patterns to the list. If you try to comment that method and start the application, the start will fail.

PR #160, opened 2023-03-06
[CHK-1191] feat(handlers): enqueue refund event after sending user KO receipt

giovanniberti:
#### List of Changes

* expire transaction then enqueue refund event after sending user KO receipt

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Depends on #159

PR #162, opened 2023-03-09
[CHK-1178] refactor(closepayment): send refund request event instead of closepayment event when getting error or KO from closePayment

giovanniberti:
#### List of Changes

* send refund request event instead of closure after bad or KO `closePayment`
* save CLOSURE_ERROR instead of CLOSURE event when getting KO or unrecoverable Nodo error on `closePayment`

#### Types of changes

#### Checklist:

Depends on #160


src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
```diff
@@ -297,26 +297,53 @@ private ClosePaymentRequestV2Dto.OutcomeEnum authorizationResultToOutcomeV2(
...
-                .filter(e ->
-                // Closed event sent on the queue only if the transaction was previously
-                // authorized and the Nodo response outcome is KO
-                TransactionClosureData.Outcome.KO.equals(e.getData().getResponseOutcome())
-                        && AuthorizationResultDto.OK.equals(authorizationResult)
+        return Mono.just(closureOutcomeEvent)
+                .filter(
+                        e -> e.fold(
+                                closureErrorEvent -> true,
+                                // Closed event sent on the queue only if the transaction was previously
```
pietro-tota, line 308:
```suggestion
                                // Refund requested event sent on the queue only if the transaction was previously
```

src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
```diff
@@ -297,26 +297,53 @@ private ClosePaymentRequestV2Dto.OutcomeEnum authorizationResultToOutcomeV2(
...
+                                            closureEvent.getTransactionId()
+                                    );
+                                    return closureEvent.getTransactionId();
+                                }
+                        )
+                )
+                .flatMap(transactionId -> {
+                    TransactionRefundRequestedEvent refundRequestedEvent = new TransactionRefundRequestedEvent(
+                            transactionId,
+                            new TransactionRefundedData(TransactionStatusDto.CLOSED)
```
pietro-tota, line 337: here the previous state can be both `CLOSURE_ERROR` or `CLOSED`  depending on the TaskEither being left or right, right?

giovanniberti: You're right, fixing it

giovanniberti: Fixed in a20f88f

pietro-tota (dismissed):
🚀

pietro-tota (approved):

PR #164, opened 2023-03-14
CHK-1150-fix-decode-transactionId

Fernando-Granato:
#### List of Changes
Handled http status 400 when decoding transactionId. 

#### Motivation and Context
Handled http status 400 when decoding transactionId. The service responded with 500

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/utils/UUIDUtils.java
+++ b/src/main/java/it/pagopa/transactions/utils/UUIDUtils.java
@@ -1,8 +1,12 @@
 package it.pagopa.transactions.utils;
 
+import it.pagopa.transactions.exceptions.InvalidRequestException;
 import org.apache.commons.codec.binary.Base64;
 import org.springframework.stereotype.Component;
+import reactor.core.publisher.Mono;
 
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


src/main/java/it/pagopa/transactions/utils/UUIDUtils.java
```diff
@@ -16,9 +20,14 @@ public String uuidToBase64(UUID uuid) {
...
-        ByteBuffer bb = ByteBuffer.wrap(bytes);
-        return new UUID(bb.getLong(), bb.getLong());
+    public Mono<UUID> uuidFromBase64(String str) {
+        try {
+            byte[] bytes = Base64.decodeBase64(str);
+            ByteBuffer bb = ByteBuffer.wrap(bytes);
+            return Mono.just(new UUID(bb.getLong(), bb.getLong()));
+        } catch (BufferUnderflowException e) {
+            return Mono.error(new InvalidRequestException("Error while decode transactionId"));
+        }
```
giovanniberti, line 30: Given that it's not an asynchronous operation is it right to return `Mono`?

Fernando-Granato: You are right, I tried to solve it using an Either let me know what you think 😉

giovanniberti (commented):
Add a test for an invalid byte array and it's on 🚀

src/test/java/it/pagopa/transactions/utils/UUIDUtilsTest.java
```diff
@@ -27,4 +28,12 @@ void shouldEncodeAndDecodeBase64OfUUID() {
 
         assertEquals(uuid, uuidFromBase64.get());
     }
+
+    @Test
+    void shouldDecodeBase64OfUUIDError() {
+        String wrongUuid = "xxxx";
+        Either<InvalidRequestException, UUID> uuidFromBase64 = uuidUtils.uuidFromBase64(wrongUuid);
+        assertEquals(InvalidRequestException.class, uuidFromBase64.getLeft().getClass());
```
giovanniberti, line 36: This line is not actually needed as it is guaranteed from the type signature, a `assertTrue(uuidFromBase64.isLeft())` should suffice

Fernando-Granato: you are right !!!! 👀

giovanniberti (approved):
🚀

PR #176, opened 2023-03-27
[CHK-1325] update to nodo activateV2

ciuffagianluca:
Adds and handles activateV2 soap call

#### List of Changes

Swap all activate nodo with activateV2.
Add test for new invocation.
Remove old activate invocations.

#### Motivation and Context

In order to get from nodo info about the trasfer list of a single rptId it is mandatory to update the acrivate nodo soap invocation. This is required also for making the most of GEC functionalities

Added junit test
Test locally

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -130,7 +129,8 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
                                                                                                 .getEcommerceFiscalCode(),
                                                                                         nodoOperations
                                                                                                 .generateRandomStringToIdempotencyKey()
-                                                                                )
+                                                                                ),
+                                                                                new ArrayList<>(5)
```
infantesimone, line 133: why 5? better a constant with a meaningful name, also you need to update https://github.com/pagopa/pagopa-ecommerce-payment-requests-service/blob/main/src/main/kotlin/it/pagopa/ecommerce/payment/requests/repositories/PaymentRequestInfo.kt

infantesimone (changes requested):

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
ciuffagianluca: This is the max size of transferListItem. Since we know this value and it is quite small, we can set it to the max value, to get optimal performance from array list, as show here https://medium.com/@malith.jayasinghe/a-performance-evaluation-of-java-arraylist-f097582b5c4d . In particular this is the conclusion _"It is clear from the from the results (considering the add operation of ArrayList) that if the required maximum capacity of the ArrayList is known, we can get the optimal performance (both average latency and throughput) by specifying the initial capacity to final size final size (or using ensureCapacity as required)"_ 
To avoid memory waste, we can do an optimization like setting the initial capacity to 3 items or 1 till the cart won't be in exercise. When the cart will be real, we can tune this parameter. What do you think?

Speaking about payment-request I'm not so sure that we must do this editing in that side, since, before activation, that list will be null or empty. What do you think about this too?

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
```diff
@@ -100,7 +99,16 @@ private Mono<PaymentRequestInfo> nodoActivationForNM3PaymentRequest(
...
                                 response.getPaymentToken(),
-                                new IdempotencyKey(idempotencyKey)
+                                new IdempotencyKey(idempotencyKey),
+                                response.getTransferList().getTransfer().stream()
+                                        .map(
+                                                transfer -> new PaymentTransferInfo(
+                                                        transfer.getFiscalCodePA(),
+                                                        transfer.getRichiestaMarcaDaBollo() != null,
+                                                        EuroUtils.euroToEuroCents(transfer.getTransferAmount()),
+                                                        null // TODO il valore non è reperibile sulla struttura dati
```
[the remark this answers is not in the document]
infantesimone: ok, so waiting SANP 3.4.0 specs :-)

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
infantesimone: ok I agree for the optimizations, but why are you talking about a cart? `cart` and `transfertList` are decoupled

infantesimone (approved):

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
ciuffagianluca: You're right, I delete confusing consideration about cart. I was wrong.

infantesimone (approved):
just a code smell "Remove this unused import 'java.util.ArrayList'", simply to fix :-)

infantesimone (approved):

PR #203, opened 2023-04-27
CHK-1428-management-rrn

Fernando-Granato:
dependOn https://github.com/pagopa/pagopa-ecommerce-commons/pull/75

#### List of Changes
Added during **PATCH** the saving of the rrn both in the view and into transactionAuthorizationEvent.

#### Motivation and Context
Saved in PATCH phase the RRN field value for VPOS payments. The field is optional
The value has been saved both on the transaction view and in the authorization completion event, to allow recovery during the call to the closePayment and sendPaymentResult for sending the mail

The tests were carried out locally

#### Types of changes

#### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <java.version>17</java.version>
         <spring-cloud-azure.version>4.0.0</spring-cloud-azure.version>
         <jacoco.version>0.8.8</jacoco.version>
-        <pagopa-ecommerce-commons.version>0.11.0</pagopa-ecommerce-commons.version>
+        <pagopa-ecommerce-commons.version>0.12.0</pagopa-ecommerce-commons.version>
         <spotless.version>2.28.0</spotless.version>
     </properties>
     <dependencies>
@@ -401,8 +401,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>branch</scmVersionType>
-                    <scmVersion>CHK-1427-save-rrn-information</scmVersion>
+                    <scmVersionType>tag</scmVersionType>
+                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
                     <goals>install -DskipTests</goals>
                 </configuration>
                 <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -401,8 +401,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>CHK-1427-save-rrn-information</scmVersion>
                     <goals>install -DskipTests</goals>
                 </configuration>
                 <executions>
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
@@ -11,6 +11,8 @@ import it.pagopa.ecommerce.commons.generated.server.model.TransactionStatusDto;
 import it.pagopa.ecommerce.commons.repositories.PaymentRequestsInfoRepository;
 import it.pagopa.generated.ecommerce.nodo.v2.dto.ClosePaymentRequestV2Dto;
 import it.pagopa.generated.ecommerce.nodo.v2.dto.ClosePaymentResponseDto;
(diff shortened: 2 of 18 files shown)


pom.xml
```diff
@@ -401,8 +401,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
```
Fernando-Granato, line 405: Restore this after merge https://github.com/pagopa/pagopa-ecommerce-commons/pull/75

src/main/java/it/pagopa/transactions/utils/AuthRequestDataUtils.java
```diff
@@ -0,0 +1,38 @@
...
+    public record AuthRequestData(
+            String authorizationCode,
+            String outcome,
+            String rrn
+    ) {
+
+    }
+
+    public AuthRequestData extract(UpdateAuthorizationRequestDto updateAuthorizationRequest) {
+        AuthRequestData result = null;
```
pietro-tota, line 24:
```suggestion
        AuthRequestData result;
```

src/main/java/it/pagopa/transactions/utils/AuthRequestDataUtils.java
```diff
@@ -0,0 +1,38 @@
...
+        switch (updateAuthorizationRequest.getOutcomeGateway()) {
+            case OutcomeVposGatewayDto t -> {
+                result = new AuthRequestData(t.getAuthorizationCode(),t.getOutcome().toString(),t.getRrn());
+            }
+            case OutcomeXpayGatewayDto t -> {
+                result = new AuthRequestData(t.getAuthorizationCode(),t.getOutcome().toString(), null);
+            }
+            default ->
+                    throw new IllegalStateException("Unexpected value: " + updateAuthorizationRequest.getOutcomeGateway());
```
pietro-tota, line 33: What about use here `it.pagopa.transactions.exceptions.InvalidRequestException` so that we will return a 400 bad request response (see controller exception handler related section)?

src/main/java/it/pagopa/transactions/utils/AuthRequestDataUtils.java
```diff
@@ -0,0 +1,38 @@
...
+    @ValueObject
+    public record AuthRequestData(
+            String authorizationCode,
+            String outcome,
+            String rrn
+    ) {
+
+    }
+
+    public AuthRequestData extract(UpdateAuthorizationRequestDto updateAuthorizationRequest) {
```
pietro-tota, line 23: Bikeshedding: call this method `from` ?

src/main/java/it/pagopa/transactions/commands/handlers/TransactionSendClosureHandler.java
```diff
@@ -114,6 +117,8 @@ public Mono<Either<TransactionClosureErrorEvent, TransactionEvent<TransactionClo
                 .flatMap(tx -> {
                     UpdateAuthorizationRequestDto updateAuthorizationRequestDto = command.getData()
                             .updateAuthorizationRequest();
+                    AuthRequestDataUtils.AuthRequestData authRequestDataExtracted = authRequestDataUtils
```
pietro-tota, line 120:
```suggestion
                    AuthRequestDataUtils.AuthRequestData authRequestData = authRequestDataUtils
```

pietro-tota: 🚲

pietro-tota (commented):
LGTM.
Just some comments.
Please resolve also some code smells (unused imports and so on)

src/main/java/it/pagopa/transactions/utils/AuthRequestDataUtils.java
```diff
@@ -0,0 +1,39 @@
...
+import it.pagopa.generated.transactions.server.model.UpdateAuthorizationRequestDto;
+import it.pagopa.transactions.exceptions.InvalidRequestException;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class AuthRequestDataUtils {
+
+    @ValueObject
```
infantesimone, line 15: why @valueObject? is it a value object according to DDD? https://domaincentric.net/blog/ddd-building-blocks-in-php-value-object#:~:text=Value%20Object%20is%20an%20object,model%20behaviour%20of%20your%20application.

infantesimone: in addition, does it make sense to have a util class for this? within the transactions domain? so in package `it/pagopa/transactions/utils` ?

infantesimone (changes requested):

src/main/java/it/pagopa/transactions/utils/AuthRequestDataUtils.java
infantesimone: hopefully we will have only one gateway in the future so it sounds good to have this class so when we have only one gateway let's delete this class.

infantesimone (dismissed):

infantesimone (approved):

PR #221, opened 2023-05-17
[CHK-1510] fix: remove null-value json properties in node requests

MarcoAGreco:
#### List of Changes

- Remove null-value json properties in node requests
- Add `JsonInclude.Include.NON_NULL` rule to jackson mapper to avoid adding null-value json properties
#### Motivation and Context

- Align to Nodo web-services requirements

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/configurations/WebClientsConfig.java
+++ b/src/main/java/it/pagopa/transactions/configurations/WebClientsConfig.java
@@ -1,5 +1,6 @@
 package it.pagopa.transactions.configurations;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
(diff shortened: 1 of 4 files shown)


src/test/java/it/pagopa/transactions/client/NodeForPspClientTest.java
```diff
@@ -220,6 +230,51 @@ void shouldReturnOKClosePaymentResponse() {
         assertThat(clientResponse.getOutcome()).isEqualTo(closePaymentResponse.getOutcome());
     }
 
+    @Captor
+    ArgumentCaptor<Publisher<ClosePaymentRequestV2Dto>> bodyCaptor;
+
+    @Test
+    void shouldReturnOKClosePaymentResponseAdditionalInfo() {
```
pietro-tota, line 237: I would rather add two tests here on the serialized ClosePaymentRequest string.
One test where additionalPaymentInformations RRN is valued and another one were is not and check that RRN is serialized in the first and not in the second one.

Maybe for let this being testable the used ObjectMapper used to build Webclient can be moved into a producer method inside WebClientsConfig class

src/main/java/it/pagopa/transactions/configurations/WebClientsConfig.java
```diff
@@ -44,6 +45,7 @@ public WebClient nodoWebClient(
             ObjectMapper mapper = new ObjectMapper();
             mapper.registerModule(new JavaTimeModule());
             mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+            mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
pietro-tota, line 48: Remember to apply this fix also into event-dispatcher module for close payment retry requests

src/test/java/it/pagopa/transactions/client/NodeForPspClientTest.java
MarcoAGreco: Thanks, done with 88c9bb9 👍

pietro-tota (approved):
🚀

PR #231, opened 2023-05-23
CHK-1301-refund_requested

Fernando-Granato:
#### List of Changes
Added a new projectionHandler for handling view update at **REFUND_REQUESTED**.
Updated transaction refund logic, specifically:
- In case of an unrecoverable error of the closePayment request you have to perform a transaction refund and update to view to **REFUND_REQUESTED**.
- In case of a recoverable error of the closePayment request you have to add the event in the retry queue and update the view state to **CLOSURE_ERROR**.
- In case of closePayment with a KO by the node if the transaction was authorized by the **PGS** a refund request must be made and the view updated to **REFUND_REQUESTED**.
-  In case of closePayment with a KO by the node if the transaction was not authorized by the **PGS** update the view to **UNAUTHORIZED**..
- In case of closePayment OK by the node you have to update the view with **CLOSED** status.

#### Motivation and Context
The goal of the activity was to manage the close payment response so that the view was updated with the correct state.

The tests were carried out in local with ecommerce-local.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/projections/handlers/RefundRequestProjectionHandler.java
+++ b/src/main/java/it/pagopa/transactions/projections/handlers/RefundRequestProjectionHandler.java
@@ -2,13 +2,16 @@ package it.pagopa.transactions.projections.handlers;
 
 import it.pagopa.ecommerce.commons.documents.v1.Transaction;
 import it.pagopa.ecommerce.commons.documents.v1.TransactionRefundRequestedEvent;
-import it.pagopa.ecommerce.commons.documents.v1.TransactionUserCanceledEvent;
 import it.pagopa.ecommerce.commons.generated.server.model.TransactionStatusDto;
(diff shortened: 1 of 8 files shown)


src/main/java/it/pagopa/transactions/projections/handlers/RefundRequestProjectionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.stereotype.Component;
+import reactor.core.publisher.Mono;
+
+@Component
+@Slf4j
+public class RefundRequestProjectionHandler
+        implements ProjectionHandler<TransactionRefundRequestedEvent, Mono<Transaction>> {
+
+    @Autowired
+    private TransactionsViewRepository transactionsViewRepository;
```
pietro-tota, line 19: Use constructor injection

pietro-tota (approved):
🚀

PR #240, opened 2023-05-31
feat: span transaction with left payment token time

pietro-tota:
Depends on https://github.com/pagopa/pagopa-ecommerce-commons/pull/88

#### List of Changes

1. Add custom elk span creation during new transaction  processing.

This span is created programmatically only when a transaction is cached for an already activated payment token.
Into custom span are present two labels:

- paymentToken --> valued with transaction payment token
- paymentTokenLeftTimeSec -> numeric label valued with payment token validity left time in seconds 

2. Move PaymentRequestInfo save after Nodo successful payment notice activation 

#### Motivation and Context
1. Those modifications are needed in order to track transactions activated close to the Nodo payment token expiration
2. Previously PaymentRequestInfo object were saved for every activation, also for those with a cache hit, causing expiration to be updated for every cache hit. With this modification Redis cache is upgraded only after Nodo activatePaymentNotice response and left as is for every subsequent cache hit, making cache leave at least as payment token duration after successful Nodo activation

Test in DEV environment

#### Screenshots (if appropriate):
Example dashboard with sample query that take into account all subsequent activation for a given payment token
Screenshot 2023-05-31 alle 16 37 17

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
+++ b/src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
@@ -224,28 +224,34 @@ public class TransactionActivateHandler
 
     private void traceRepeatedActivation(PaymentRequestInfo paymentRequestInfo) {
         String transactionActivationDateString = paymentRequestInfo.activationDate();
-        Duration paymentTokenValidityTimeLeft = Duration.ofSeconds(paymentTokenTimeout);
+
         if (transactionActivationDateString != null) {
             ZonedDateTime transactionActivation = ZonedDateTime.parse(transactionActivationDateString);
             ZonedDateTime paymentTokenValidityEnd = transactionActivation
                     .plus(Duration.ofSeconds(paymentTokenTimeout));
-            paymentTokenValidityTimeLeft = Duration.between(ZonedDateTime.now(), paymentTokenValidityEnd);
-        }
-        log.info(
-                "PaymentRequestInfo cache hit for {} with valid paymentToken {}. Validity left time: {}",
-                paymentRequestInfo.id(),
-                paymentRequestInfo.paymentToken(),
-                paymentTokenValidityTimeLeft
-        );
-        co.elastic.apm.api.Span span = ElasticApm.currentTransaction()
-                .startSpan("transaction", "activation", "activation cached");
-        try {
-            span
-                    .setName("Transaction re-activated")
-                    .setLabel("paymentToken", paymentRequestInfo.paymentToken())
-                    .setLabel("paymentTokenLeftTimeSec", paymentTokenValidityTimeLeft.getSeconds());
-        } finally {
-            span.end();
+            Duration paymentTokenValidityTimeLeft = Duration.between(ZonedDateTime.now(), paymentTokenValidityEnd);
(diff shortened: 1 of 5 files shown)


Fernando-Granato (approved):

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -230,6 +222,34 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
         );
     }
 
+    private void spanCustomTransaction(PaymentRequestInfo paymentRequestInfo) {
```
giovanniberti, line 225:
```suggestion
    private void traceRepeatedActivation(PaymentRequestInfo paymentRequestInfo) {
```

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -230,6 +222,34 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
...
+        String transactionActivationDateString = paymentRequestInfo.activationDate();
+        Duration paymentTokenValidityTimeLeft = Duration.ofSeconds(paymentTokenTimeout);
```
giovanniberti, line 227: Isn't this the the payment token validity time? (i.e. should be the same as the first activation, not less)

pietro-tota: The paymentTokenTimeout, valued as class attribute, contains the `payment.token.validity` configured value

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
pietro-tota: Fixed with https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/240/commits/0da79d3734a91df7b44ea40e704686c6bfe6e30e

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
giovanniberti: Ok, I would remove `left` in the name then, what do you think?

src/main/java/it/pagopa/transactions/commands/handlers/TransactionActivateHandler.java
```diff
@@ -230,6 +222,45 @@ public Mono<Tuple2<Mono<TransactionActivatedEvent>, String>> handle(
...
+                );
+            } else {
+                log.error(
+                        "Cannot trace repeated transaction activation for {} with payment token: {}, missing transaction activation date",
+                        paymentRequestInfo.id(),
+                        paymentRequestInfo.paymentToken()
+                );
+                span.captureException(
+                        new IllegalArgumentException(
+                                "Invalid transaction activation date null for rptId: %s"
```
giovanniberti, line 253:
```suggestion
                                "Null transaction activation date for rptId %s in repeated activation"
```

giovanniberti (approved):
🚀

PR #244, opened 2023-06-12
feat: added clientId query param  into close payment request

Fernando-Granato:
#### List of Changes
Added into **closePayment** request new query param **_ecomm_**.

#### Motivation and Context
The change was made to add a new query param to the close payment request to the node. 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
+++ b/src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
@@ -26,6 +26,7 @@ import javax.xml.bind.JAXBElement;
 @Slf4j
 public class NodeForPspClient {
 
+    private static final String CLIENT_ID_CLOSE_PAYMENT = "ecomm";
     @Autowired
     @Qualifier("nodoWebClient")
     private WebClient nodoWebClient;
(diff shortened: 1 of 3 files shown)


src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
```diff
@@ -26,6 +26,7 @@
 @Slf4j
 public class NodeForPspClient {
 
+    private static final String CLOSE_PAYMENT_CLIENT_ID = "ecomm";
```
pietro-tota, line 29: I would put this as configuration parameter valued with ecomm, what do you think?

pietro-tota (dismissed):

src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
Fernando-Granato: I believe that the parameter could be of little use considering that we will not change it so easily.
What do you think ? 
However, if we wanted we could parameterize it.

pietro-tota: Yes, maybe you're right, nevermind :D

pietro-tota (approved):

PR #256, opened 2023-06-21
[CHK-1663] Mask all digits string from 3 to 20 chars

ciuffagianluca:
Edit mask pattern to mask all group of digits that can identify a pan

#### List of Changes

Add safe pattern list to avoid masking timestamp data

#### Motivation and Context

Security of data

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/utils/EcsEncoderLogMasker.java
+++ b/src/main/java/it/pagopa/transactions/utils/EcsEncoderLogMasker.java
@@ -2,7 +2,12 @@ package it.pagopa.transactions.utils;
 
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import co.elastic.logging.logback.EcsEncoder;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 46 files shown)


src/main/java/it/pagopa/transactions/utils/EcsEncoderLogMasker.java
```diff
@@ -29,13 +36,34 @@ public byte[] encode(ILoggingEvent event) {
 
     private String maskMessage(String message) {
         StringBuilder sb = new StringBuilder(message);
+        ArrayList<List<Integer>> safeIndexes = new ArrayList<>();
```
pietro-tota, line 39: An idea to avoid add safe pattern processing (to skip log timestamps) could be the following: 
here we are masking the whole log message that is in json format
(ex: `{"@timestamp":"2023-06-23T07:22:20.892Z","log.level": "INFO","message":"Transaction initialized for rptId: 77777777777311116723749670106","ecs.version": "1.2.0","service.name":"pagopa-ecommerce-transactions-service","service.version":"0.19.13","service.environment": │
│`). 
Maybe here can be worth to mask only the message field content so that the timestamp exclusion can be skipped altogether as done here https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/4f5658bc3ecd5cc3d072227385ba969b9643484f/src/test/java/it/pagopa/transactions/utils/EcsEncoderLogMaskerTests.java#L58-L60.

So masking, then, can be applied to the log message field only masking that field with the configured regex leaving other log fields intact.

What do you think?

ciuffagianluca: It's a good idea, i'm investigating to find a way to manipulate the JsonObject

ciuffagianluca: @pietro-tota I did as you suggested!

src/main/java/it/pagopa/transactions/utils/EcsEncoderLogMasker.java
```diff
@@ -15,6 +19,12 @@ public class EcsEncoderLogMasker extends EcsEncoder {
     private Pattern patternString;
     private final List<String> maskPatterns = new ArrayList<>();
 
+    private ObjectMapper objectMapper;
+
+    public EcsEncoderLogMasker() {
```
pietro-tota, line 24: Add package private constructor with ObjectMapper as input parameter so that you don't have to use reflections into tests

pietro-tota (approved):

PR #267, opened 2023-07-11
[CHK-1692] fix(activation): propagate verification due date

pietro-tota:
#### List of Changes
Propagate dueDate saved during verification process into cache informations saved after Nodo activate call.

#### Motivation and Context
In the current transaction activation flow, due date is overridden by transaction service that set this field forcibly to null.
This way, a new verification performed after transaction activation, that will not communicate with Nodo because of cache hit, will return a response without the dueDate field valued

Junit tests

#### Types of changes

#### Checklist:


ciuffagianluca (approved):
Well done! 🚀

PR #286, opened 2023-07-31
fix(circuit-breaker): remove misconfigured circuit breaker

pietro-tota:
#### List of Changes
Remove circuit breaker from api's that have not external dependency other than Azure resources.
Api's from which circuit breaker configuration has been removed are: 

- user cancel
- authorization status update
- add user receipt

#### Motivation and Context

Those api's, in fact, has only Azure as downstream dependency (such as storage queues, cosmos db etc) a no any external service so there is no need for add circuit breaker.
Moreover, the previous configuration cause add user receipt requests to not be processed successfully because of the opened circuit breaker that was wrongly configured as the one used for transaction activation causing avoidable errors returned to the caller (since no downstream dependency was effectively down)

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/services/TransactionsService.java
+++ b/src/main/java/it/pagopa/transactions/services/TransactionsService.java
@@ -193,7 +193,6 @@ public class TransactionsService {
                 );
     }
 
-    @CircuitBreaker(name = "transactions-backend")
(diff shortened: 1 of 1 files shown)


giovanniberti (approved):

ciuffagianluca (approved):
🚀

PR #297, opened 2023-08-29
feat(npg): CHK 1831 retrieve bin from payment method service

Fernando-Granato:
#### List of Changes
Added in case of authorization request to be forwarded to NPG the logic for card data retrieval using payment-methods.

#### Motivation and Context
The change was made to manage the authorization flow for NPG.

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/services/TransactionsService.java
```diff
@@ -776,14 +793,13 @@ NewTransactionResponseDto.ClientIdEnum convertClientId(
                 ).orElseThrow(() -> new InvalidRequestException("Null value as input origin"));
     }
 
-    private String extractBinFromPan(RequestAuthorizationRequestDto requestAuthorizationRequestDto) {
+    private Mono<Optional<String>> extractBinFromPan(RequestAuthorizationRequestDto requestAuthorizationRequestDto) {
```
infantesimone, line 796: rename this method?

infantesimone: the point is that `bin` is not retrieved only from the `pan`, so better rename method name

Fernando-Granato: Fixed with d232dd40676e53de7e9bbbf5080c3e2f99d49f69!

infantesimone (approved):

infantesimone:
to check integration test fails

PR #311, opened 2023-09-18
CHK-1905 : feat(jwt): added claim ORDER_ID into jwt

Fernando-Granato:
#### List of Changes
👉🏻 Added optional field orderId into POST transaction body.
👉🏻 Added optional claim orderId into JWT.
👉🏻 Added into NPG flow orderId into transactionActivatedEvent

#### Motivation and Context
The change was made to save the orderId value in the case of a request for the NPG flow.

The tests were carried out with ecommerce-local

#### Types of changes

#### Checklist:


api-spec/transactions-api.yaml
```diff
@@ -362,6 +362,10 @@ components:
           description: Cart identifier provided by creditor institution
           type: string
           example: idCartFromCreditorInstitution
+        orderId:
```
pietro-tota, line 365: I think that we can create an api `v2` version with mandatory field `orderId`.
This way we can add also other fields that will be required for NPG integration, such as `correlationId` and so on, making those fields mandatory.
What do you think? @infantesimone

pietro-tota (approved):

PR #328, opened 2023-09-25
feat(nodo): add apy-key for auth nodo ws

infantesimone:
#### Types of changes

#### Checklist:


infantesimone:
depends on https://github.com/pagopa/pagopa-ecommerce-local/pull/98

helm/values-dev.yaml
```diff
@@ -76,7 +76,7 @@ microservice-chart:
     REDIS_PORT: "6380"
     REDIS_SSL_ENABLED: "true"
     NODO_HOSTNAME: https://api.dev.platform.pagopa.it
-    NODE_FOR_PSP_URI: /nodo/node-for-psp/v1
+    NODE_FOR_PSP_URI: /nodo-auth/node-for-psp/v1
```
pietro-tota, line 79: Missing `NODO_NODEFORPSP_API_KEY` and `NODO_CLOSEPAYMENT_API_KEY` env secrets?

src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
```diff
@@ -32,6 +32,12 @@ public class NodeForPspClient {
     @Value("${nodo.nodeforpsp.uri}")
     private String nodoPerPspUri;
 
+    @Value("${nodo.nodeforpsp.apikey}")
```
pietro-tota, line 35: Maybe here we can refactor this class to use constructor injection? what do you think? this way we can avoid reflection to set those api key into test class

infantesimone: right, but better to do it in a dedicated PR

helm/values-dev.yaml
infantesimone: fixed

src/main/java/it/pagopa/transactions/client/NodeForPspClient.java
```diff
@@ -81,10 +88,11 @@ public Mono<ClosePaymentResponseDto> closePaymentV2(ClosePaymentRequestV2Dto req
         );
         return nodoWebClient.post()
                 .uri(
-                        uriBuilder -> uriBuilder.path("/nodo/nodo-per-pm/v2/closepayment")
+                        uriBuilder -> uriBuilder.path("/nodo-auth/nodo-per-pm/v2/closepayment")
```
pietro-tota, line 91: I would parameterized (maybe in another PR) this path in order to change it's value upgrading helm instead of modifying code. What do you think?

pietro-tota (approved):
Well done, just some trivial issue. 🚀

pietro-tota:
Please remember to make a PR for environments alignment against https://github.com/pagopa/pagopa-ecommerce-local

PR #396, opened 2024-01-11
[CHK-2321] feat(checkout-redirect): add secret that holds PSPs' API keys

giovanniberti:
#### List of Changes

 * add environment secret to hold PSP API keys for Checkout Redirect
 * add Checkout Redirect PSP API keys as Spring properties
 * add `CheckoutRedirectPspApiKeysConfig` bean to hold Checkout Redirect PSP API keys

#### Motivation and Context

This PR adds to the application environment a secret that holds the API keys needed to do the first call to PSPs integrated with Checkout Redirect.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Depends on pagopa/pagopa-ecommerce-commons#166

PR #397, opened 2024-01-11
[CHK-2319] -  checkout redirect implementation api client configuration

pietro-tota:
Depends on https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/396

#### List of Changes
Add Checkout Redirect api client configuration.
Since b2b Redirection api's will be implemented by every PSP this configuration handles PSP to api client mapping returning the api client configured for each PSP dinamically

#### Motivation and Context
This implementation handle api client for PSP dynamic generation

Junit tests + tests with eCommerce local

#### Types of changes

#### Checklist:


api-spec/redirect/redirect-api.yaml
```diff
@@ -0,0 +1,218 @@
+openapi: 3.0.0
```
infantesimone, line 1: is this openapi used for client generation? could we insert a folder in the file path that indicates this? for example `client/openapi/*.yaml` or `openapi/client/*.yaml`

infantesimone (dismissed):

api-spec/redirect/redirect-api.yaml
pietro-tota: Yes I like it. Can we leave api-spec as root folder or maybe is preferable to move it out?

pietro-tota: Fixed with 99dd8836a5ea433dadf7c77a54c5de8b9a0475e6

api-spec/client/openapi/redirect/redirect-api.yaml
```diff
@@ -0,0 +1,218 @@
...
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/RedirectUrlRequest'
+  securitySchemes:
+    PspApiKeyAuth:
+      type: apiKey
+      in: header
+      name: x-api-key
```
infantesimone, line 217: TO DO: add refs to "nuova connettività" (and update api header `ocp-apim-subscription-key`)

pietro-tota: Ok, seems legit. Just one doubt, this way PSP implementation will expect api key to be present into ocp-apim-subscription-key header? Or this modification affers to api exposed by us to PSP (aka the outcome request)?

giovanniberti: Shouldn't this header be only on the API interface exposed on `pagopa-api`?

pietro-tota: @infantesimone merging this pr as is for the api key header name, will be updated in the future after integration clarification

giovanniberti (approved):

PR #412, opened 2024-01-29
feat(redirect outcome): add patch handling for REDIRECT payment flow

pietro-tota:
Infra related [PR](https://github.com/pagopa/pagopa-infra/pull/1688) to update api specs

#### List of Changes
Add PATCH auth details for redirect authorization payment flows.
This detail maps the PSP outcome received informations in order to properly process authorization outcome result.
In this PR have been add also the following authorization outcome request fields validation:

- input psp id is the same as the one selected during transaction flow
- input pspTransactionId is the same as the one returned by the PSP into retrieve url response
- request is received before authorization outcome timeout expiration (as per the retrieve url response timeout field)

#### Motivation and Context
Those modifications are needed in order to properly handle redirect authorization outcome PSP api calls

Junit tests + local tests with eCommerce local

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionUpdateAuthorizationHandler.java
```diff
@@ -58,42 +60,143 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionUpdateAuthorizationComman
...
+
+    /**
+     * This method performs validation check against received update authorization
+     * request
+     *
+     * @param command the handler command
+     * @return an empty mono if request was validated successfully otherwise a mono
+     *         error with request validation failure details
+     */
+    private Mono<Void> isUpdateTransactionRequestValid(TransactionUpdateAuthorizationCommand command) {
```
giovanniberti, line 120: What about moving this validation logic inside `TransactionUpdateRequesetCommand`'s constructor instead?

giovanniberti (approved):

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionUpdateAuthorizationHandler.java
pietro-tota: But here event store access is needed in order to reduce events and retrieve authorization data.
Also this is more of a semantic authorization request validation rather than a syntactic one (so a bean validation), performing checks also on authorization request timestamp and so on.
Idk, I would rather leave this validation logic here, or it can be moved to a utility class.
I see this piece of code as a pre-processing and maybe leave it here can help finding validation logic.
What do you think?

pietro-tota: Fixed with https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/412/commits/52d16a8accb30f8d01a0005a88b44eebe08546a4

giovanniberti (dismissed):

giovanniberti (approved):

PR #421, opened 2024-02-06
[CHK-2334] fix(deps): mark `netty-resolver-dns-native-macos` dependency for test only as it causes memleak in prod

giovanniberti:
#### List of Changes

 * marked dependency `netty-resolver-dns-native-macos` with test scope

#### Motivation and Context

The dependency was needed for macOS on ARM arch and was included unconditionally, but it causes a memory leak of non-heap memory in Linux hosts.

Deployment in dev environment and monitoring on memory.

#### Types of changes

#### Checklist:


Fernando-Granato (approved):
🚀🚀

PR #407, opened 2024-01-25
[CHK-2368] fix(redirect): update redirect API spec

giovanniberti:
#### List of Changes

 * switch `idPaymentMethod` and `paymentMethod` in api spec
 * change semantics of `paymentMethod` field in Redirect URL retrieval API
 * change domain of `idPaymentMethod` field
 * add more examples to URL Retrieval API

#### Motivation and Context

This PR updates the incoming Checkout Redirect API spec with the latest requirements

#### How Has This Been Tested?

Local unit tests

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Depends on #403

PR #432, opened 2024-02-13
CHK-2479 - handle payment method and payment method id redirect retrieve url api body fields

pietro-tota:
Depends on https://github.com/pagopa/pagopa-ecommerce-local/pull/144

#### List of Changes
Handle idPaymentMethod and paymentMethod fields population into redirection retrieve url api calls

#### Motivation and Context
This PR restore the feature that was developed in this [PR](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/407) 

Junit tests

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -83,6 +82,41 @@ public class PaymentGatewayClient {
...
+        static RedirectPaymentMethodId fromPaymentTypeCode(String paymentTypeCode) {
+            RedirectPaymentMethodId converted = lookupMap.get(paymentTypeCode);
+            if (converted == null) {
+                throw new InvalidRequestException(
+                        "Unmanaged payment method with type code: [%s]".formatted(paymentTypeCode)
+                );
+            }
+            return converted;
+        }
+    }
```
giovanniberti, line 118: Isn't this the same as `valueOf()`? (except for the exception thrown)

giovanniberti (approved):

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
pietro-tota: Yes, I put it just to thrown the Invalid request exception to return a 400 bad request instead of an Internal Server Error.

PR #436, opened 2024-02-16
[CHK-2470] session token into outcome URL

Fernando-Granato:
#### List of Changes
Added jwt token for outcome page useful for fe to poll downstream of an APM payment with wallet.
Modified the logic to generate the 'outcome url.

#### Motivation and Context
The change was made to allow successful completion of an APM payment with the wallet.

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -672,4 +665,61 @@ private Either<CheckoutRedirectConfigurationException, URI> getRedirectUrlForPsp
...
+                                                                        outcomeToken
+                                                                )
+                                                        )
+                                                )
+                                )
+
+                );
+    }
+
+    private URI generateNpgOutcomeUrl(
```
pietro-tota, line 707: 🚲  this method is used by both npg and redirect payment flows. What about refactoring it to 
```suggestion
    private URI generateOutcomeUrl(
```

Fernando-Granato: Fixed with c294728dd00f565352271c99e453544d4a954731

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -543,107 +536,107 @@ public Mono<RedirectUrlResponseDto> requestRedirectUrlAuthorization(
...
-                                                "Unhandled error performing redirection URL api call for PSP: [%s]"
-                                                        .formatted(pspId),
-                                                null
-                                        );
-                                    }
-                                }
+        return new JwtTokenUtils()
+                .generateToken(
+                        npgNotificationSigningKey,
+                        npgJwtKeyValidityTime,
```
pietro-tota, line 542:
```suggestion
                        npgNotificationSigningKey,
                        npgJwtKeyValidityTime,
```
This token is the one that will be used by eCommerce f.e. for polling the transaction status in the outcome page. Please use the eCommerce signing key here

Fernando-Granato: Excellent observation fixed with 56689e716478674f1d3421ef9eb15b8ebe407ecc

pietro-tota (approved):

PR #463, opened 2024-03-25
fix(redirect): add psp id to url mapping

pietro-tota:
#### List of Changes
Add psp id to redirect url mapping (as key) in order to discriminate redirect urls also by psp id

#### Motivation and Context
This modification is needed since redirect url mapping needs to be discriminated also by psp id (same payment method exposed by multiple psps)

Junit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
+++ b/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
@@ -589,6 +589,7 @@ public class PaymentGatewayClient {
                                     .idPaymentMethod(idPaymentMethod.toString())
                                     .paName(null);// optional
                             Either<RedirectConfigurationException, URI> pspConfiguredUrl = getRedirectUrlForPsp(
+                                    authorizationData.pspId(),
                                     authorizationData.paymentTypeCode()
                             );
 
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -91,9 +91,9 @@ SESSION_URL_NOTIFICATION_URL="http://localhost/ecommerce/npg/notifications/v1/se
 JWT_NPG_NOTIFICATION_SECRET=ODMzNUZBNTZENDg3NTYyREUyNDhGNDdCRUZDNzI3NDMzMzQwNTFEREZGQ0MyQzA5Mjc1RjY2NTQ1NDk5MDMxNzU5NDc0NUVFMTdDMDhGNzk4Q0Q3RENFMEJBODE1NURDREExNEY2Mzk4QzFEMTU0NTExNjUyMEExMzMwMTdDMDk
 
 JWT_NPG_NOTIFICATION_VALIDITY_TIME=900
-REDIRECT_PAYMENT_TYPE_CODE_LIST=RPIC,RABC
+REDIRECT_PAYMENT_TYPE_CODE_LIST=REDIRECT-RBPR
```
giovanniberti, line 94: [nitpick]
What do you think about removing `REDIRECT-` and using payment type codes directly?

giovanniberti (dismissed):

.env.example
pietro-tota: The aim of this pr is to modify redirect url mapping as pspId-paymentTypeCode. For mocked environment the specific psp id configured as redirect (gec mock and so on have id REDIRECT). 
So this is now the new mapping :D

pietro-tota: Maybe here we can remove the "psp" part for the configured keys but then we have less control against configured expected keys

.devops/code-review-pipelines.yml
```diff
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: main
+      ref: CHK-2598-handle-url-for-psp-and-type-code
```
```suggestion
      ref: main
```
TODO to be replaced after dependent ecommerce-local pr merge

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -661,13 +662,17 @@ private String encodeMdcFields(AuthorizationRequestData authorizationData) {
...
+                                                                             String paymentTypeCode
+    ) {
+        String urlKey = "%s-%s".formatted(pspId, paymentTypeCode);
+        if (redirectBeApiCallUriMap.containsKey(urlKey)) {
+            return Either.right(redirectBeApiCallUriMap.get(urlKey));
         } else {
             return Either.left(
                     new RedirectConfigurationException(
-                            "Missing key for payment type code: [%s]".formatted(paymentTypeCode),
+                            "Missing key for redirect return url [%s]".formatted(urlKey),
```
pietro-tota, line 675:
```suggestion
                            "Missing key for redirect return url with key: [%s]".formatted(urlKey),
```

giovanniberti (approved):

infantesimone (approved):

PR #489, opened 2024-05-09
[CHK-2828] Check user in transactions operations

ciuffagianluca:
Add check users for the following operations:

- GET transaction
- DELETE transaction
- POST authorization-request

#### List of Changes

APIs definition accept an header field `x-user-id` used to be checked against the transaction document user id value. This check is applied only for v2 document and when the userId is set. 
So we can distinguish 2 case and relative sub-cases:

1.  `V1` document: check is done against the value of the xUserId passed from the query. If it is null the check is ok. Otherwise (not accepted case by design, `V1` document hasn't the field `userId` ) the result is cut.
2. `V2` document: check is done against the value of the `xUserId` passed from the query:
     - if it is null and the document field is null, the result is accepted
     - if is in not null, the check is done from xUserId.toString() against document.userId field. If the values are equal. the result is accepted.      

If the field is null and the transaction is from v2 domains, it will be checked the null value for userId in the document.

#### Motivation and Context

Add security check for authenticated transactions (appIO)

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
+++ b/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
@@ -7,8 +7,7 @@ import reactor.core.publisher.Mono;
 public interface TransactionsViewRepository extends ReactiveCrudRepository<BaseTransactionView, String> {
     Mono<BaseTransactionView> findByTransactionId(String transactionId);
 
-    Mono<BaseTransactionView> findByTransactionIdAndUserId(
-                                                           String id,
-                                                           String userId
-    );
+    /*
+     * Mono<BaseTransactionView> findById( String id, String userId );
+     */
 }
--- a/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
+++ b/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
@@ -8,12 +8,4 @@ import reactor.core.publisher.Mono;
 public interface TransactionsViewRepository extends ReactiveCrudRepository<BaseTransactionView, String> {
     Mono<BaseTransactionView> findByTransactionId(String transactionId);
 
-    @Query(
-            value = "{ 'transactionId' : ?0 , 'userId' : ?1}"
-    )
-    Mono<BaseTransactionView> findByTransactionIdAndUserId(
-                                                           String transactionId,
-                                                           String userId
-    );
-
 }
--- a/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
+++ b/src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
@@ -7,7 +7,7 @@ import reactor.core.publisher.Mono;
 public interface TransactionsViewRepository extends ReactiveCrudRepository<BaseTransactionView, String> {
     Mono<BaseTransactionView> findByTransactionId(String transactionId);
 
-    Mono<BaseTransactionView> findByIdAndUserId(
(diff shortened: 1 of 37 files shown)


src/main/java/it/pagopa/transactions/repositories/TransactionsViewRepository.java
```diff
@@ -1,9 +1,11 @@
 package it.pagopa.transactions.repositories;
 
 import it.pagopa.ecommerce.commons.documents.BaseTransactionView;
+import org.springframework.data.mongodb.repository.Query;
 import org.springframework.data.repository.reactive.ReactiveCrudRepository;
 import reactor.core.publisher.Mono;
 
 public interface TransactionsViewRepository extends ReactiveCrudRepository<BaseTransactionView, String> {
     Mono<BaseTransactionView> findByTransactionId(String transactionId);
+
```
giovanniberti, line 10: Stray empty line
```suggestion
```

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
```diff
@@ -711,6 +712,15 @@ public Mono<RequestAuthorizationResponseDto> requestTransactionAuthorization(
                 );
     }
 
+    private Mono<BaseTransactionView> getBaseTransactionView(String transactionId, UUID xUserId) {
+        return transactionsViewRepository.findById(transactionId)
+                    .filter(transactionDocument -> switch (transactionDocument) {
+                        case it.pagopa.ecommerce.commons.documents.v1.Transaction ignored -> true;
+                        case it.pagopa.ecommerce.commons.documents.v2.Transaction t -> xUserId == null ? t.getUserId() == null : t.getUserId().equals(xUserId.toString());
```
giovanniberti, line 719: Alternative:
```suggestion
                        case it.pagopa.ecommerce.commons.documents.v2.Transaction t -> Objects.equals(xUserId.toString(), t.getUserId());
```

petretiandrea (dismissed):

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
```diff
@@ -711,6 +712,15 @@ public Mono<RequestAuthorizationResponseDto> requestTransactionAuthorization(
...
+                        default -> throw new NotImplementedException("Handling for transaction document: [%s] not implemented yet".formatted(transactionDocument.getClass()));
```
giovanniberti, line 720:
```suggestion
                        default -> throw new NotImplementedException("Handling for transaction document version: [%s] not implemented yet".formatted(transactionDocument.getClass()));
```

src/test/java/it/pagopa/transactions/services/v2/TransactionServiceTests.java
```diff
@@ -283,15 +283,85 @@ void getTransactionReturnsTransactionDataOriginProvided() {
...
 
-        StepVerifier.create(transactionsServiceV1.getTransactionInfo(TRANSACTION_ID))
+        StepVerifier
+                .create(transactionsServiceV1.getTransactionInfo(TRANSACTION_ID, UUID.fromString(USER_ID)))
                 .expectNext(expected)
                 .verifyComplete();
     }
 
+    @Test
+    void getTransactionReturnsTransactionV2() {
```
giovanniberti, line 297:
```suggestion
    void getTransactionReturnsTransactionV2WithNullUserId() {
```

giovanniberti (approved):

giovanniberti (approved):

PR #502, opened 2024-05-24
fix: add span to circuit breaker open

petretiandrea:
This PR add a opentelemetry span when the circuit breaker is OPEN.

#### List of Changes
- add span to opened circuit breaker error handler

#### Motivation and Context
In many cases, during NPG PROD sessions, elastic reports trace without any span info when the response error is caused by an opened circuit breaker. It's useful to distinguish when an error is caused by circuit breaker or not.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
+++ b/src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
@@ -62,7 +62,7 @@ public class TransactionsController implements TransactionsApi {
         }
     )
     public Mono<ResponseEntity<ProblemJsonDto>> openStateHandler(CallNotPermittedException error) {
-        log.error("Error - OPEN circuit breaker");
+        log.error("Error - OPEN circuit breaker", error);
         return Mono.just(
                 new ResponseEntity<>(
                         new ProblemJsonDto()
--- a/src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
+++ b/src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
@@ -75,10 +75,7 @@ public class TransactionsController implements TransactionsApi {
                 ignored -> openTelemetryUtils.addErrorSpanWithAttributes(
                         OpenTelemetryUtils.CIRCUIT_BREAKER_OPEN_SPAN_NAME
                                 .formatted(error.getCausingCircuitBreakerName()),
-                        Attributes.of(
-                                OpenTelemetryUtils.CIRCUIT_BREAKER_OPEN_NAME_ATTRIBUTE_KEY,
-                                error.getCausingCircuitBreakerName()
-                        )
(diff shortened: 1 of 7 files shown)


src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
```diff
@@ -51,12 +53,15 @@ public class TransactionsController implements TransactionsApi {
...
+    @Autowired
+    private OpenTelemetryUtils openTelemetryUtils;
+
     @ExceptionHandler(
         {
                 CallNotPermittedException.class
         }
     )
-    public Mono<ResponseEntity<ProblemJsonDto>> openStateHandler() {
+    public Mono<ResponseEntity<ProblemJsonDto>> openStateHandler(CallNotPermittedException error) {
```
pietro-tota, line 64: please add exception also to error log in order to log stacktrace into logs

petretiandrea: Solved by https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/502/commits/9549d6c61c741e1c1eda455abc490570a15a12ec

src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
```diff
@@ -66,6 +71,15 @@ public Mono<ResponseEntity<ProblemJsonDto>> openStateHandler() {
                                 .detail("Upstream service temporary unavailable. Open circuit breaker."),
                         HttpStatus.BAD_GATEWAY
                 )
+        ).doOnNext(
+                ignored -> openTelemetryUtils.addErrorSpanWithAttributes(
+                        OpenTelemetryUtils.CIRCUIT_BREAKER_OPEN_SPAN_NAME
+                                .formatted(error.getCausingCircuitBreakerName()),
+                        Attributes.of(
+                                OpenTelemetryUtils.CIRCUIT_BREAKER_OPEN_NAME_ATTRIBUTE_KEY,
```
pietro-tota, line 79: redundant information? circuit breaker name is already present into error span name, since this will be used only to trace circuit breaker errors I would not add this attribute (and avoid add a new span attribute to the indexed ones on elastic). what do you think?

petretiandrea: Yes, you're right, I don't think we need to make an indexed search on circuit breaker name, so I remove the attribute

petretiandrea: Removed here https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/502/commits/69181d2294821fc0f7325cf2d857c8fc3a3d10be

pietro-tota (approved):

PR #505, opened 2024-05-28
[CHK-2882] [CHK-2884] fix(auth-request) remove requestTransactionAuthorization circuit breaker

ciuffagianluca:
In order to avoid request failing for temporary timeout from NPG, circuit breaker is temporary removed for this api.
A better analysis of NPG response code and confirmPayment behavior will allow to reintroduce a more smart circuit breaker for this api.

Update npg auth reqeust timeout to 120s after data report analysis on authorization request time

#### Types of changes

#### Checklist:


petretiandrea (dismissed):

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
```diff
@@ -451,7 +451,6 @@ public Mono<Void> cancelTransaction(String transactionId, UUID xUserId) {
 
     }
 
-    @CircuitBreaker(name = "transactions-backend")
     @Retry(name = "requestTransactionAuthorization")
```
pietro-tota, line 454: what about remove also retry handler here? for 5xx error codes retry will already be performed by f.e.
What do you think?

ciuffagianluca: We decide to let the retry here for the time we will switch from PGS to NPG and back

PR #525, opened 2024-06-13
[CHK-2938] feat: add `v2.1` API with tokenized email in `POST /transactions`

giovanniberti:
#### List of Changes
Adds new API version with tokenized email

#### Motivation and Context
In previous API versions the emails are transmitted in clear text.

Unit test

#### Types of changes

#### Checklist:

Depends on pagopa/pagopa-ecommerce-commons#207
Depends on pagopa/pagopa-ecommerce-local#167


pietro-tota (dismissed):
🚀 well done!

api-spec/v3/transactions-api.yaml
```diff
@@ -0,0 +1,718 @@
+openapi: 3.0.0
```
pietro-tota, line 1: what about take into account the POST transaction body detail `feature` to be included into this v3 api version? I'm talking about the fact to introduce a detail section into POST transaction to distinguish data for an NPG and a Redirect transaction so that we can remove the hardcoded "orderId" into policies.
It can be just prepared here and v2 policies can call the service v3 version with the "npg detail" so that in the future we can take into account also the detail diversification without introducing a v4 version (and a version misalignment between infra version and service one)

pom.xml
```diff
@@ -530,8 +555,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>mock-confidential-data-manager</scmVersion>
                     <goals>install -DskipTests</goals>
```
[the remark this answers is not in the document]
giovanniberti: Fixed in f6c0c1c1

api-spec/v3/transactions-api.yaml
pietro-tota: yes, this could be a solution. My only concern was about creating a mismatch between apim exposed api and internal api, a situation for which apim v3 POST transaction calls the internal v4 and so on making version matching a bit difficult messy

pietro-tota (approved):

api-spec/v3/transactions-api.yaml
infantesimone: I agree with the comment, this v3 could be the opportunity to fix the orderId field. 
if we were to create a new external v3 to the app, should we create a v4?

giovanniberti: Renamed to `v2.1`, `v3` addressed in https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/537

src/main/java/it/pagopa/transactions/controllers/v2_1/TransactionsController.java
```diff
@@ -0,0 +1,278 @@
...
+
+    @Warmup
+    public void postNewTransactionWarmupMethod() {
+        IntStream.range(0, 3).forEach(
+                idx -> {
+                    log.info("Performing warmup iteration: {}", idx);
+                    NewTransactionResponseDto newTransactionResponseDto = WebClient
+                            .create()
+                            .post()
+                            .uri("http://localhost:8080/v3/transactions")
```
pietro-tota, line 256:
```suggestion
                            .uri("http://localhost:8080/v2.1/transactions")
```
this should be v 2.1, right?

src/main/java/it/pagopa/transactions/services/v2_1/TransactionsService.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+@Service(TransactionsService.QUALIFIER_NAME)
+@Slf4j
+public class TransactionsService {
+
+    public static final String QUALIFIER_NAME = "TransactionsServiceV3";
```
pietro-tota, line 29:
```suggestion
    public static final String QUALIFIER_NAME = "TransactionsServiceV2.1";
```

src/main/java/it/pagopa/transactions/controllers/v2_1/TransactionsController.java
giovanniberti: Yep, fixed with latest push

src/main/java/it/pagopa/transactions/services/v2_1/TransactionsService.java
giovanniberti: Changed in e37d13e3

pietro-tota (approved):
Well done 🚀

.devops/code-review-pipelines.yml
```diff
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: main
+      ref: CHK-2938-tokenized-email
```
giovanniberti, line 15: To revert after merge of https://github.com/pagopa/pagopa-ecommerce-local/pull/167

pietro-tota (approved):

PR #551, opened 2024-07-23
[CHK-3055] Add span for send payment result monitoring

ciuffagianluca:
Add span to track send payment result in kibana

#### List of Changes

Moved inside the handler the succes case to track for the send payment result processing in order to have more info about the transaction on which the send payment result is about.
Introducing some info in AlreadyProcessedException to track the info about error flow in send payment result
Introducing new Processing exception to track edge case of send payment result sent when transaction is in ACTIVATED state

#### Motivation and Context

Improve ELK monitoring

Tests have been run in dev env

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestUserReceiptHandler.java
```diff
@@ -73,7 +83,30 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionAddUserReceiptCommand com
...
+                                    .getTransactionGatewayAuthorizationRequestedData()).getWalletInfo() != null;
+                    return Mono.error(
+                            new AlreadyProcessedException(
+                                    t.getTransactionId(),
+                                    t.getTransactionAuthorizationRequestData().getPspId(),
+                                    t.getTransactionAuthorizationRequestData().getPaymentTypeCode(),
+                                    t.getClientId().name(),
+                                    isWalletPayment,
+                                    // TODO Qui ci va il codice di risposta della send payment result ? L'error code
+                                    // sarà sempre empty() ?
```
giovanniberti, line 102: `Option.empty()` is ok

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestUserReceiptHandler.java
```diff
@@ -104,62 +137,111 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionAddUserReceiptCommand com
...
+                                            language,
+                                            addUserReceiptRequestDto.getPaymentDate().toZonedDateTime().toString()
                                     )
-                                            .doOnNext(
-                                                    queueResponse -> log.info(
+                            ),
+                            tx
+                    );
+                }).flatMap(
+                        event_tx -> userReceiptAddedEventRepository.save(event_tx.getT1())
```
giovanniberti, line 193: If you want to avoid doing `event_tx.getT1()` and so on you could refactor this into

```suggestion
                        TupleUtils.function((event, tx) -> userReceiptAddedEventRepository.save(event)
```

src/main/java/it/pagopa/transactions/configurations/UpdateTransactionStatusTracerUtilsConfig.java
```diff
@@ -1,5 +1,6 @@
 package it.pagopa.transactions.configurations;
 
+import io.opentelemetry.api.trace.Tracer;
```
giovanniberti, line 3: Stray import?

src/test/java/it/pagopa/transactions/controllers/v1/TransactionsControllerTest.java
```diff
@@ -1184,20 +1185,117 @@ void shouldReturnUnprocessableEntityInSendPaymentResultForUncaughtError() {
...
+        String contextPath = "auth-requests";
         /*
-         * UpdateTransactionStatusTracerUtils.NodoStatusUpdate expectedStatusUpdateInfo
-         * = new UpdateTransactionStatusTracerUtils.NodoStatusUpdate(
+         * UpdateTransactionStatusTracerUtils.PaymentGatewayStatusUpdate
+         * expectedStatusUpdateInfo = new
+         * UpdateTransactionStatusTracerUtils.PaymentGatewayStatusUpdate(
          * UpdateTransactionStatusTracerUtils.UpdateTransactionStatusOutcome.
-         * INVALID_REQUEST );
+         * INVALID_REQUEST, expectedTrigger, Optional.empty(), Optional.empty() );
```
giovanniberti, line 1285: Stray comment?

src/test/java/it/pagopa/transactions/utils/UpdateTransactionStatusTracerUtilsTest.java
```diff
@@ -0,0 +1 @@
+
```
giovanniberti, line 1: empty file? :)

giovanniberti (approved):

src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
```diff
@@ -264,6 +287,56 @@ public Mono<ResponseEntity<AddUserReceiptResponseDto>> addUserReceipt(
                 );
     }
 
+    /**
+     * This method maps input throwable to proper {@link UpdateTransactionStatusTracerUtils.UpdateTransactionStatusOutcome} enumeration
```
pietro-tota, line 291:
```suggestion
     * This method maps input throwable to proper {@link UpdateTransactionStatusTracerUtils.SendPaymentResultOutcomeInfo} enumeration
```

src/main/java/it/pagopa/transactions/exceptions/ProcessingErrorException.java
```diff
@@ -0,0 +1,9 @@
+package it.pagopa.transactions.exceptions;
+
+public class ProcessingErrorException extends RuntimeException {
```
pietro-tota, line 3: This exception is mapped to 422 HTTP response code by code [here](https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/0bce94e59031680a7dec2d6ed9ed80928483bebf/src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java#L501-L522). Is this wanted or is a side-effect?

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestUserReceiptHandler.java
```diff
@@ -74,38 +74,56 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionAddUserReceiptCommand com
                 );
 
         Mono<it.pagopa.ecommerce.commons.domain.v2.TransactionClosed> alreadyProcessedError = transaction
```
giovanniberti, line 76: Just food for thought, maybe better for another PR: given that this chain is quite long and filled with conditionals, what do you think about making a method that computes the result?

giovanniberti (approved):
:rocket:

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestUserReceiptHandler.java
ciuffagianluca: I'll open an issue to track this

src/main/java/it/pagopa/transactions/exceptions/ProcessingErrorException.java
ciuffagianluca: It is wanted. 
In this exception I basically want to track this edge case (send payment result for a transction without an authorization, really not possibile). Is this a little bit different from the Already processed exception: here we map the fact that we don't expect this call from nodo simply because the transaction isn't yet authorized. So the 422 response code is ok.

PR #557, opened 2024-07-31
[CHK-3119] feat: map `WISP_REDIRECT` client id when creating transactions to `CHECKOUT_CART`

giovanniberti:
#### Motivation and Context

This PR includes the groundwork from pagopa/pagopa-ecommerce-commons#214
The returned field `clientId` is mapped to `CHECKOUT_CART` in order to make the new `WISP` client backward compatible with the current API

#### Types of changes

#### Checklist:

Depends on pagopa/pagopa-ecommerce-commons#214

### DIFF
--- a/api-spec/v2.1/transactions-api.yaml
+++ b/api-spec/v2.1/transactions-api.yaml
@@ -278,7 +278,7 @@ components:
         - IO
         - CHECKOUT
         - CHECKOUT_CART
-        - WISP
+        - WISP_REDIRECT
     UserId:
       description: unique user identifier
       type: string
--- a/api-spec/v2.1/transactions-api.yaml
+++ b/api-spec/v2.1/transactions-api.yaml
@@ -278,6 +278,7 @@ components:
         - IO
         - CHECKOUT
         - CHECKOUT_CART
+        - WISP
     UserId:
       description: unique user identifier
       type: string
--- a/pom.xml
+++ b/pom.xml
@@ -559,8 +559,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -559,8 +559,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>CHK-3104-wisp-dismantling-client-id</scmVersion>
```
giovanniberti, line 563: TODO fix after depending PR merge

src/main/java/it/pagopa/transactions/services/v2_1/TransactionsService.java
```diff
@@ -160,19 +160,22 @@ private Mono<NewTransactionResponseDto> projectActivatedEvent(
...
-                                                                  String clientId
+                                                                  Transaction.ClientId clientId
     ) {
         return Optional.ofNullable(clientId).filter(Objects::nonNull)
                 .map(
                         value -> {
                             try {
-                                return NewTransactionResponseDto.ClientIdEnum.fromValue(value);
+                                return switch (clientId) {
+                                    case WISP -> NewTransactionResponseDto.ClientIdEnum.CHECKOUT_CART;
```
pietro-tota, line 176:
```suggestion
                                    case WISP_REDIRECT -> NewTransactionResponseDto.ClientIdEnum.CHECKOUT_CART;
```

giovanniberti: Fixed in f879b87f

api-spec/v2.1/transactions-api.yaml
```diff
@@ -278,6 +278,7 @@ components:
         - IO
         - CHECKOUT
         - CHECKOUT_CART
+        - WISP
```
petretiandrea, line 281: In order to avoid breaking changes we had decided not to add clientId or am I wrong?

giovanniberti: This is used only in `POST /transactions`, so it's not breaking
It's still needed as a change to support the use case of activating a transaction without calling verify on it beforehand

petretiandrea (approved):

PR #558, opened 2024-08-01
feat: [CHK-3109] Handle payment notice IUV

petretiandrea:
Depends on https://github.com/pagopa/pagopa-ecommerce-payment-requests-service/pull/111

This PR introduces handling for IUV notice number. Also address some critical point:
- expose only effective client to rest API
- use only IO or CHECKOUT as touchpoints to GET /payment-methods
- use only effective clients as touchpoints for calculate fees
- returns `creditorReferenceId` during transaction activation (only for `WISP_REDIRECT`)

#### List of Changes
- get and save `creditorReferenceId` from Nodo activation during POST /transactions
- remap clientId to IO or CHECKOUT when getting payment method -> [see](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-b658409e3962e249360ebea4b1a737fed620afab5ff5cd3064edec397fa05336R62)
- extract `creditorReferenceId` only for `WISP_REDIRECT` client -> [see](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-d82b5a1bc392730c24e3b03a7909ba49dec26a6c3b77327d679ac29b7b492e3aR9)
- return `creditorReferenceId` to POST /transactions when client is `WISP_REDIRECT` -> [v2.1](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-7e2b904cadc7a364448287b1811ee8948cc58d365e5ac9738901f823e63307a5R139) and [v2](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-abe906d3c9a2319ffd7565282f30839392df85e70946283960506e3ae9e9aadcR147)
- calculating fee using effective touchpoint (e.g. `WISP_REDIRECT` becomes `CHECKOUT_CART`) -> [see](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-c633b723f784d3debc9f0af7170b95214b0b3684c3b2774924952a3bfa0e351eR521)
- hiding internal client and return only the effective client (e.g. `WISP_REDIRECT` becomes `CHECKOUT_CART`) -> [see](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/files#diff-c633b723f784d3debc9f0af7170b95214b0b3684c3b2774924952a3bfa0e351eR420)

#### Types of changes

#### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <java.version>17</java.version>
         <spring-cloud-azure.version>4.0.0</spring-cloud-azure.version>
         <jacoco.version>0.8.8</jacoco.version>
-        <pagopa-ecommerce-commons.version>1.23.1</pagopa-ecommerce-commons.version>
+        <pagopa-ecommerce-commons.version>1.24.0</pagopa-ecommerce-commons.version>
         <spotless.version>2.28.0</spotless.version>
         <ecs-logging-version>1.5.0</ecs-logging-version>
         <mock-web-server.version>4.11.0</mock-web-server.version>
@@ -559,8 +559,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>branch</scmVersionType>
-                    <scmVersion>CHK-3107-iuv-payment-notice</scmVersion>
+                    <scmVersionType>tag</scmVersionType>
+                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
                     <goals>install -DskipTests</goals>
                 </configuration>
                 <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -559,8 +559,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>CHK-3107-iuv-payment-notice</scmVersion>
                     <goals>install -DskipTests</goals>
                 </configuration>
                 <executions>
--- a/src/main/java/it/pagopa/transactions/client/EcommercePaymentMethodsClient.java
+++ b/src/main/java/it/pagopa/transactions/client/EcommercePaymentMethodsClient.java
@@ -1,5 +1,6 @@
 package it.pagopa.transactions.client;
 
+import it.pagopa.ecommerce.commons.documents.v2.Transaction;
 import it.pagopa.generated.ecommerce.paymentmethods.v1.dto.PatchSessionRequestDto;
 import it.pagopa.generated.ecommerce.paymentmethods.v1.dto.PaymentMethodResponseDto;
 import it.pagopa.generated.ecommerce.paymentmethods.v1.dto.SessionPaymentMethodResponseDto;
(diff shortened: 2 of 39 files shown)


pom.xml
```diff
@@ -559,8 +559,8 @@
                 <configuration>
                     <skipCheckoutIfExists>false</skipCheckoutIfExists>
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
-                    <scmVersionType>tag</scmVersionType>
-                    <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
```
petretiandrea, line 562: Bump to next ecommerce-common version

src/main/java/it/pagopa/transactions/client/EcommercePaymentMethodsClient.java
```diff
@@ -57,7 +58,12 @@ public Mono<PaymentMethodResponseDto> getPaymentMethod(
                                                            String paymentMethodId,
                                                            String xClientId
     ) {
-        return ecommercePaymentMethodsWebClientV1.getPaymentMethod(paymentMethodId, xClientId)
+        // payment methods only support CHECKOUT and IO.
```
pietro-tota, line 61: maybe here we can add a task to trace the CHECKOUT_CART and CHECKOUT methods discrimination since GEC allow bundle configuration for cart and single payment as independent configurations

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
```diff
@@ -163,7 +163,8 @@ private Mono<PaymentRequestInfo> nodoActivationForNM3PaymentRequest(
                                                         transfer.getTransferCategory()
                                                 )
                                         ).toList(),
-                                isAllCCP(response, allCCPOnTransferIbanEnabled)
+                                isAllCCP(response, allCCPOnTransferIbanEnabled),
+                                response.getCreditorReferenceId()
```
pietro-tota, line 167: I would add a Nodo semantic response validation here. For WISP redirect transactions we expect that `creditorReferenceId` would be always valued, a null value for those cases will lead to an error since we would not have a value to be shown to the client (checkout) and to be sent into the email body.

So for wisp redirect transaction this value should be present and not null here (it's nullable into specs) and for null response value I would throw an `InvalidNodoResponseException` here

petretiandrea: Solved in https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/commits/b55bcbb47b9d4b75ec1e2a97af89d1aeab4eed45 by wrapping nodo activate payment

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
```diff
@@ -58,6 +59,42 @@ public NodoOperations(
...
+                                        )
+                                )
+                        );
+                    } else {
+                        return Mono.just(paymentRequestInfo);
+                    }
+                });
+    }
+
     public Mono<PaymentRequestInfo> activatePaymentRequest(
```
giovanniberti, line 98: Can we make this overload private? (If the overload that should be used is the one with `clientId`)

giovanniberti (approved):
LGTM 🚀

pom.xml
petretiandrea: Solved by https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/commits/be7f777206043f365ad0f02b16691aa6662e39a1

src/main/java/it/pagopa/transactions/client/EcommercePaymentMethodsClient.java
petretiandrea: We decide not to distinguish `CHECKOUT_CART` from `CHECKOUT`, at least for now, right?

src/main/java/it/pagopa/transactions/utils/NodoOperations.java
petretiandrea: Sure, done in https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/558/commits/6d10998f3e66aca1c9fe9bb279f794d18a4624ca

giovanniberti (approved):

pietro-tota (approved):

PR #560, opened 2024-08-02
[CHK-3094] fix(redirect): truncate `paName` field to 70 chars

giovanniberti:
#### List of Changes

 * truncate `paName` field to 70 chars

#### Motivation and Context

This change fixes a discrepancy between Checkout Redirect `paName` length (max length 70) and Nodo's `activate` `companyName` field (max length 140).
Now, a `companyName` field that is longer than 70 chars gets truncated to 67 chars plus an ellipsis `...`.

#### Types of changes

#### Checklist:

Depends on pagopa/pagopa-mock-ec#67

### DIFF
--- a/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
+++ b/src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
@@ -738,7 +738,7 @@ public class PaymentGatewayClient {
         if (paName == null || paName.length() <= 70) {
             return paName;
         } else {
-            return paName.substring(0, 69) + "…";
+            return paName.substring(0, 67) + "...";
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -733,6 +734,14 @@ public Mono<RedirectUrlResponseDto> requestRedirectUrlAuthorization(
 
     }
 
+    private String shortenRedirectPaName(@Nullable String paName) {
+        if (paName == null || paName.length() <= 70) {
+            return paName;
+        } else {
+            return paName.substring(0, 69) + "…";
```
pietro-tota, line 741:
```suggestion
            return paName.substring(0, 67) + "...";
```

pietro-tota: I know but I would avoid use special characters that will go to external entities... maybe we can truncate to 70 characters without adding any trailing character

giovanniberti: Fixed in e23847b7

pietro-tota (approved):

PR #563, opened 2024-08-08
feat: [CHK-3139] Expose http metrics

petretiandrea:
#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -1,3 +1,12 @@
+management.metrics.enable:
+    all: false
+    application: false # reports startup time, useful for future comparison
+    http.server.requests: true # reports http request counts
+    resilience4j.circuitbreaker: true # reports circuit breaker status
+    jvm.memory: true # reports heap status
+    jvm.gc: true # reports garbage collector metrics, useful to detect too long pauses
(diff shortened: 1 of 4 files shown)


infantesimone:
This service is the core of eCommerce, the performance is very important.
These metrics shouldn't have (important) impact on the service, but has it been evaluated? tuning frequency and number of metrics collected?

src/main/resources/application.yml
```diff
@@ -1,3 +1,12 @@
+management.metrics.enable:
+    all: false
+    application: true # reports startup time, useful for future comparison
+    http.server.requests: true # reports http request counts
+    resilience4j.circuitbreaker: true # reports circuit breaker status
+    jvm.memory: true # reports heap status
+    jvm.gc: true # reports garbage collector metrics, useful to detect too long pauses
+    mongodb.driver: true # reports useful mongo statistics, like waiting queue
```
pietro-tota, line 8: I would expose only requests count metrics disabling others. those can be enabled in the future as needed. 
What do you think?

petretiandrea: Yes, I'm agree, the purpose was to get more data during testing, so now, I will disable useless metrics :)

PR #574, opened 2024-09-17
fix: avoid multiple auth requests for same transaction

pietro-tota:
#### List of Changes
Make auth request performable only once per transaction id, regarding the api call outcome, using Redis locking mechanism

#### Motivation and Context
Order id is calculated on the fly for APM and wallet transactions.
In case of timeout receiving NPG response or if client cut the connection a subsequent retry will generate a new order id that will be used in NPG transaction.
This finally results in an order id being lost (not saved into DB) for an authorization request that in traced into NPG systems but not into eCommerce.
Using the same orderId is not an option since multiple api call to NPG with the same orderId will result in multiple transaction being processed NPG side with the same transaction id -> this will result in a transaction receiving multiple outcomes (CANCELED after the first attempt expiration) and the second attempt one.
So the solution that reconciliate with eCommerce flow is to limit auth requests per transaction to one and only one, despite the api processing outcome (since in case of any error there could be established a communication with NPG).
For implementation simplicity and reliability it was chosen to put a locking mechanism upon POST auth requests api call to just limit to one the performable api calls per transaction, avoiding multiple authorization being performed for the same transaction

Junit tests + local test using eCommerce local

#### Types of changes

#### Checklist:


src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
```diff
@@ -148,6 +152,32 @@ public Mono<ResponseEntity<RequestAuthorizationResponseDto>> requestTransactionA
...
+                    ExclusiveLockDocument lockDocument = new ExclusiveLockDocument(
+                            "POST-auth-request-%s".formatted(transactionId),
+                            "transactions-service"
+                    );
+                    // fix CHK-3222: auth request operations are not idempotent NPG side, so this
+                    // lock prevents multiple auth request to be performed for a single transaction
+                    // (for timeouts scenario and so on). the lock duration have been set to the
+                    // payment token validity time in order to make this api call performable only
+                    // once for transaction (further attempts will found transaction in expired
+                    // status and return an error)
```
giovanniberti, line 165: wording
```suggestion
                    // fix CHK-3222: auth request operations are not idempotent on the NPG side, so this
                    // lock prevents multiple auth request to be performed for a single transaction
                    // (for timeouts scenarios, etc.). The lock duration has been set to the
                    // payment token validity time in order to make this API call performable only
                    // once per transaction (further attempts will find the transaction in an expired
                    // status and return an error)
```

giovanniberti (approved):

giovanniberti (approved):

giovanniberti (approved):

src/main/java/it/pagopa/transactions/controllers/v1/TransactionsController.java
```diff
@@ -148,6 +152,32 @@ public Mono<ResponseEntity<RequestAuthorizationResponseDto>> requestTransactionA
     ) {
         return requestAuthorizationRequestDto
                 .doOnNext(t -> log.info("RequestTransactionAuthorization for transactionId: [{}]", transactionId))
+                .map(updateAuthorizationRequest -> {
```
infantesimone, line 155: maybe this lock is too restrictive, if we have  a db connection fails, we will block these calls, without the beneficial effects of retry

pietro-tota: We have to distinguish between DB connections fails, if they happen reading transaction from DB or writing event to event store.

Said that we could move this locking logic right before performing gateway api call, what do you think?

pietro-tota: Yes, in exception handler, but for a DB exception how can you tell were it happens? Fetching information from db or writing? 

IMHO moving locking just before performing api calls to gateway could be a right trade-off: for any error that could happen after communication have been initiated with gateway a retry should be avoid (there could be also network errors receiving response, such as timeout, but the authorization process could already be processed gateway side)

PR #585, opened 2024-10-21
[CHK-3323] feat: append random queryparam to prevent caching

scaminati-bv:
#### List of Changes
- add **t** queryparam with timestamp in milliseconds into npg outcome url
- add **t** queryparam with timestamp in milliseconds into auth-request urls

#### Motivation and Context
The goal of this PR is to prevent _index.html_ caching by append random queryparam to redirect urls

Unit tests

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

#### Checklist:


src/test/java/it/pagopa/transactions/utils/UrlUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static Map<String, String> getParametersAsMap(String query) {
+        return getParametersAsMap(query, false);
+    }
+
+    public static Map<String, String> getParametersAsMap(
+                                                         String query,
+                                                         boolean removeRandom
+    ) {
+        if (query == null || query.isEmpty()) {
```
EmanueleBVtech, line 40: Instead of checking for `null `directly, you can use `Optional `to handle potential `null `values more gracefully.

   ```
 public static Map<String, String> getParametersAsMap(String query, boolean removeRandom) {
        return Optional.ofNullable(query)
                       .map(q -> Arrays.stream(q.split("&"))
                                       .map(param -> param.split("="))
                                       .filter(pair -> !removeRandom || !pair[0].equals("t"))
                                       .collect(Collectors.toMap(
                                           pair -> pair[0],
                                           pair -> pair.length > 1 ? pair[1] : ""
                                       )))
                       .orElse(Map.of());
    }
```

scaminati-bv: Done here ffce41c35a3c04e52f19b7bd8478fcea89aa893e.
But in this suggestion is missing empty string check, without this a map with a wrong value would be returned.

EmanueleBVtech: you a right, try something like this?


`public static Map<String, String> getParametersAsMap(String query, boolean removeRandom) {
    return Optional.ofNullable(query)
                   .filter(q -> !q.isEmpty()) // Check if the query is not empty
                   .map(q -> Arrays.stream(q.split("&"))
                                   .map(param -> param.split("="))
                                   .filter(pair -> !removeRandom || !pair[0].equals("t"))
                                   .collect(Collectors.toMap(
                                       pair -> pair[0],
                                       pair -> pair.length > 1 ? pair[1] : ""
                                   )))
                   .orElse(Map.of());
}
`

scaminati-bv: Yes is already done in the previous commit

ciuffagianluca (approved):

PR #597, opened 2024-12-12
feat: add get transaction v2

pietro-tota:
Depends on https://github.com/pagopa/pagopa-ecommerce-commons/pull/225
Depends on https://github.com/pagopa/pagopa-ecommerce-local/pull/179

#### List of Changes
Add GET transaction V2 api call.
This api is an evolution for the v1 version where:
- it have been perform a response body refactoring to aggregate gateway and node information into response body
- add close payment error Node specific information

#### Motivation and Context
those modification are needed in order to make:
1. response body clearer
2. return to the caller touchpoint the close payment Node error additional information needed to display error messages to the user based on errors during close payment phase

junit tests + local test with eCommerce local + integrations test in dev env

#### Types of changes

#### Checklist:

### DIFF
--- a/.devops/code-review-pipelines.yml
+++ b/.devops/code-review-pipelines.yml
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: main
+      ref: CHK-3616-close-payment-errors
       endpoint: 'io-azure-devops-github-ro'
 
 stages:
--- a/.devops/code-review-pipelines.yml
+++ b/.devops/code-review-pipelines.yml
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: CHK-3616-close-payment-errors
+      ref: main
       endpoint: 'io-azure-devops-github-ro'
 
 stages:
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
         <java.version>17</java.version>
         <spring-cloud-azure.version>4.0.0</spring-cloud-azure.version>
         <jacoco.version>0.8.8</jacoco.version>
-        <pagopa-ecommerce-commons.version>1.31.1</pagopa-ecommerce-commons.version>
+        <pagopa-ecommerce-commons.version>1.32.0</pagopa-ecommerce-commons.version>
         <spotless.version>2.28.0</spotless.version>
         <ecs-logging-version>1.5.0</ecs-logging-version>
         <mock-web-server.version>4.11.0</mock-web-server.version>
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -538,6 +538,8 @@
                     <connectionUrl>scm:git:https://github.com/pagopa/pagopa-ecommerce-commons.git</connectionUrl>
                     <scmVersionType>tag</scmVersionType>
                     <scmVersion>${pagopa-ecommerce-commons.version}</scmVersion>
+                    <scmVersionType>branch</scmVersionType>
+                    <scmVersion>CHK-3614-close-payment-error</scmVersion>
```
pietro-tota, line 542:
```suggestion
```
TODO to be removed after dependent PR merge

.devops/code-review-pipelines.yml
```diff
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: main
+      ref: CHK-3616-close-payment-errors
```
```suggestion
      ref: main
```

TODO: to be replaced after dependent ecommerce-local PR merge

CianoDanilo (approved):

src/main/java/it/pagopa/transactions/services/v2/TransactionsService.java
```diff
@@ -197,4 +207,95 @@ public NewTransactionResponseDto.ClientIdEnum convertClientId(
...
                 ).orElseThrow(() -> new InvalidRequestException("Null value as input origin"));
     }
+
+    @CircuitBreaker(name = "ecommerce-db")
+    @Retry(name = "getTransactionInfo")
+    public Mono<it.pagopa.generated.transactions.v2.server.model.TransactionInfoDto> getTransactionInfo(
+                                                                                                        String transactionId,
+                                                                                                        UUID xUserId
+    ) {
+        log.info("Get Transaction Invoked with id {} ", transactionId);
```
EmanueleBVtech, line 217: Since transactionId and xUserId are not marked as nullable in this function,
maybe we should add a check for them to not be null before using them?


```   
@CircuitBreaker(name = "ecommerce-db")
@Retry(name = "getTransactionInfo")
public Mono<it.pagopa.generated.transactions.v2.server.model.TransactionInfoDto> getTransactionInfo(
    String transactionId,
    UUID xUserId
) {
    validateInputs(transactionId, xUserId); <------
    
    return getBaseTransactionView(transactionId, xUserId)
        .switchIfEmpty(Mono.error(new TransactionNotFoundException(transactionId)))
        .map(this::buildTransactionInfoDtoFromView)
        .doOnError(e -> log.error("Error retrieving transaction info for id {}: {}", transactionId, e.getMessage()));
}



private void validateInputs(String transactionId, UUID xUserId) {
    if (transactionId == null || transactionId.isEmpty()) {
        throw new IllegalArgumentException("Transaction ID cannot be null or empty");
    }
    if (xUserId == null) {
        throw new IllegalArgumentException("User ID cannot be null");
    }
}
`

pietro-tota: Input validation is already performed by the validation framework at controller level.

transactionId is set as required parameter in openapi, so a null value can never happen here

xUserId, instead, is an optional value, set only for transactions made by authenticated user (IO app). Checkout transaction are guest transactions so there is no xUserId to be set here and no filter to be made against transaction

EmanueleBVtech (approved):

CianoDanilo (approved):

EmanueleBVtech (approved):

PR #611, opened 2025-04-09
feat: add specs for patch auth request v2

pietro-tota:
Depends on https://github.com/pagopa/pagopa-ecommerce-local/pull/189

#### List of Changes
Create PATCH /transactions/{transactionId}/auth-requests V2 api version that accept the transaction id as path parameter instead of the base64 encoded version.

#### Motivation and Context
The base64 encode trick was done for the already dismantled PGS gateway flow to accomplish to a strict transaction id length constraint that is no more in place.
Removing this base64 encoding will allow to search for all api involved in a transaction by it's transaction id that will be the same for all api

junit tests + local tests using eCommerce local

#### Types of changes

#### Checklist:


.devops/code-review-pipelines.yml
```diff
@@ -12,7 +12,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local 
-      ref: main
+      ref: CHK-3992-patch-auth-request-v2
```
```suggestion
      ref: main
```

TODO: to replace after dependent pr merge

sfarralorenzo (dismissed):

EmanueleBVtech (dismissed):

sfarralorenzo (approved):

EmanueleBVtech (approved):
:rocket

PR #661, opened 2025-07-23
feat: [CHK-4354] delete write transaction view

forciniti-luigi-jakala:
#### List of Changes

- Enable/disable writing to transactions-view collection via feature flag (TRANSACTIONSVIEW_UPDATE_ENABLED)

#### Motivation and Context

This PR introduces a feature flag to control whether the service writes to the transactions-view collection. The motivation is to provide flexibility in managing updates to the view, allowing the feature to be enabled or disabled without code changes. This supports scenarios like gradual rollout, testing, or temporary disabling of view updates without impacting the main transaction processing flow.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
+++ b/src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
@@ -6,15 +6,15 @@ import io.vavr.control.Either;
 import it.pagopa.ecommerce.commons.documents.BaseTransactionEvent;
 import it.pagopa.ecommerce.commons.documents.BaseTransactionView;
 import it.pagopa.ecommerce.commons.documents.PaymentTransferInformation;
-import it.pagopa.ecommerce.commons.documents.v2.Transaction;
-import it.pagopa.ecommerce.commons.domain.v2.*;
 import it.pagopa.ecommerce.commons.documents.v2.*;
+import it.pagopa.ecommerce.commons.documents.v2.Transaction;
 import it.pagopa.ecommerce.commons.domain.v1.EmptyTransaction;
 import it.pagopa.ecommerce.commons.domain.v1.TransactionEventCode;
-import it.pagopa.ecommerce.commons.redis.templatewrappers.v2.PaymentRequestInfoRedisTemplateWrapper;
 import it.pagopa.ecommerce.commons.domain.v1.pojos.BaseTransaction;
+import it.pagopa.ecommerce.commons.domain.v2.*;
 import it.pagopa.ecommerce.commons.domain.v2.pojos.BaseTransactionWithPaymentToken;
 import it.pagopa.ecommerce.commons.generated.server.model.TransactionStatusDto;
+import it.pagopa.ecommerce.commons.redis.templatewrappers.v2.PaymentRequestInfoRedisTemplateWrapper;
 import it.pagopa.ecommerce.commons.utils.UpdateTransactionStatusTracerUtils;
 import it.pagopa.generated.ecommerce.paymentmethods.v2.dto.*;
 import it.pagopa.generated.transactions.server.model.*;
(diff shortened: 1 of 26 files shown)


forciniti-luigi-jakala:
Maybe this was already considered. We're running some tests with @Lorenzo on dev using the transactions-service with the feature flag disabled, and paying by credit card.

Right after entering the card details and clicking "Pay", we get a 404 on the POST /ecommerce/checkout/v1/transactions/:tId/auth-requests, which, looking at the code, goes directly to the transactions-view.

sfarralorenzo (dismissed):

dylantangredi-jakala:
LGTM but should we fix the trivial new sonar issues introduced on the tests? (maybe the unused import aswell from April since we are cleaning up anyway)

[New Issues introduced from July, 21st](https://sonarcloud.io/project/issues?createdAfter=2025-07-21&createdBefore=2025-07-29&issueStatuses=OPEN%2CCONFIRMED&pullRequest=661&id=pagopa_pagopa-ecommerce-transactions-service)

dylantangredi-jakala (dismissed):

ciuffagianluca (commented):
Is this the final solution for this task?
IMHO this is not a good solution, because the handle methods returns an object (the transaction document) that should be saved in the database, but, in case of `write-view-flag` disabled is only computed by the handle and not saved. 
The portion of code that invoke the handlers should have all the events to make a reduction of them and build a view of the transactions on the fly.
I try to explain it better. 
The solution implemented is one of that kind:
```
public Document handle(Event event) {
saveEvent(event)
view = computeView(event)
if(writeEnabled) {
save(view);
}
return view;
}

```

I think is better a solution of this kind

```
events = findFromEventStore()
...
if(writeViewEnabled) {
view = handle(event) // the old one
} else {
eventSaved = handle(event); //write view disabled
view = computeViewFromEvents(events, eventSaved);
}
return view;
...
```

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
```diff
@@ -145,7 +147,8 @@ public TransactionsService(
             UpdateTransactionStatusTracerUtils updateTransactionStatusTracerUtils,
             @Value("#{${npg.authorizationErrorCodeMapping}}") Map<String, String> npgAuthorizationErrorCodeMapping,
             @Value("${ecommerce.finalStates}") Set<String> ecommerceFinalStates,
-            @Value("${ecommerce.possibleFinalStates}") Set<String> ecommercePossibleFinalStates
+            @Value("${ecommerce.possibleFinalStates}") Set<String> ecommercePossibleFinalStates,
+            @Value("${transactionsview.update.enabled}") boolean transactionsviewUpdateEnabled
```
pietro-tota, line 151: to be removed, this ff is evaluated at projection handler level, right?

forciniti-luigi-jakala: You're correct

ciuffagianluca (approved):

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
forciniti-luigi-jakala: fixed in a37e9683f3631261101ecce0dfa474b9a203b3d7

src/test/java/it/pagopa/transactions/commands/handlers/v2/TransactionsActivationProjectionHandlerTest.java
```diff
@@ -24,16 +24,26 @@
...
     TransactionsViewRepository transactionsViewRepository;
 
+    @BeforeEach
+    void setUp() {
+        handler = new TransactionsActivationProjectionHandler(
+                transactionsViewRepository,
+                true
+        );
+    }
+
```
pietro-tota, line 46: avoid initialize this bean for each test, initialize it once (as done for AuthorizationProjectionHandlerTest for example)

src/test/java/it/pagopa/transactions/projections/handlers/v2/AuthorizationUpdateProjectionHandlerTest.java
```diff
@@ -304,4 +240,92 @@ void shouldSaveErrorCodeForNpgKOAuthorizationRequest() {
...
+                Transaction.ClientId.CHECKOUT,
+                transaction.getTransactionActivatedData().getIdCart(),
+                paymentTokenValidity,
+                new EmptyTransactionGatewayActivationData(),
+                TransactionTestUtils.USER_ID
+        );
+
+    }
+
+    private static Transaction getRrn(TransactionActivated transaction) {
```
pietro-tota, line 311: what is this getRrn method means? rrn stands for "retrieval reference number (or something like that)", a payment terminology to indicate a payment identifier. here this method returns a view object, leading to misunderstandings.
Please remove this method and use just one for transaction view object creation (some from pagopa-ecommerce-commons TransactionUtils class)

pietro-tota (approved):

sfarralorenzo (approved):

PR #664, opened 2025-08-04
feat: enable AOT compilation

pietro-tota:
#### Types of changes

#### Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,15 +6,16 @@ RUN apk add --no-cache git
 COPY mvnw .
 COPY .mvn .mvn
 COPY pom.xml .
+
 #validate step will execute the scm plugin to perform checkout and installation of the pagopa-commons library
-RUN ./mvnw validate -DskipTests
+RUN ./mvnw validate -DskipTests -Dmaven.site.skip=true
 RUN ./mvnw dependency:copy-dependencies
 RUN ./mvnw dependency:go-offline
 
 COPY src src
 COPY api-spec api-spec
 COPY eclipse-style.xml eclipse-style.xml
-RUN ./mvnw install -DskipTests --offline
+RUN ./mvnw compile spring-boot:process-aot install -DskipTests --offline
 RUN mkdir target/extracted && java -Djarmode=layertools -jar target/*.jar extract --destination target/extracted
 
 FROM eclipse-temurin:21-jre-alpine@sha256:8728e354e012e18310faa7f364d00185277dec741f4f6d593af6c61fc0eb15fd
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,16 +6,54 @@ RUN apk add --no-cache git
...
+else cp -R ../extracted/spring-boot-loader/* ./; \
+fi
+RUN if [ -z "$(ls -A ../extracted/snapshot-dependencies)" ]; \
+then echo "Skipped empty folder"; \
+else cp -R ../extracted/snapshot-dependencies/* ./; \
+fi
+RUN if [ -z "$(ls -A ../extracted/application)" ]; \
+then echo "Skipped empty folder"; \
+else cp -R ../extracted/application/* ./; \
+fi
```
dylantangredi-jakala, line 49: Non-blocker comment

Should we combine the 4 different RUN commands into a single RUN command using a loop? To avoid creating unnecessary layers for a single logical operation (preparing the CDS classpath) and to follow DRY principle

something like 

```
RUN for dir in dependencies spring-boot-loader snapshot-dependencies application; do \
      if [ -d "../extracted/$dir" ] && [ "$(ls -A ../extracted/$dir)" ]; then \
        cp -r ../extracted/$dir/* ./; \
      fi; \
    done
```

Dockerfile
```diff
@@ -24,17 +62,24 @@ USER user:user
...
 WORKDIR /app/
 
-ARG EXTRACTED=/workspace/app/target/extracted
+ARG EXTRACTED=/workspace/app/extracted
 #ELK Agent
 ADD --chown=user https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v2.1.0/opentelemetry-javaagent.jar .
 
-COPY --from=build --chown=user ${EXTRACTED}/dependencies/ ./
+COPY --from=optimizer --chown=user ${EXTRACTED}/dependencies/ ./
+RUN true
```
dylantangredi-jakala, line 70: Do we need to add more RUN true commands after COPY? I thought this was a practice for old docker versions that had issues with failing COPY commands, that now serves no purpose if not adding an empty layer. Can we consider removing this new one and maybe the existing ones as well since this is an optimization PR anyway? 

sources:
https://stackoverflow.com/questions/73301757/run-true-in-dockerfile
https://stackoverflow.com/questions/29762231/how-can-i-understand-the-role-of-bin-true-command-in-docker-run-command

dylantangredi-jakala (dismissed):

CianoDanilo (approved):

infantesimone (approved):

PR #675, opened 2025-08-29
[CHK-4393] Remove view find to check transaction data

ciuffagianluca:
Refactoring to remove all view findByID invocation and use the events reduction or replace with activated event find to check transaction exists, get data and check if user is present and its the same passed by jwt token

#### List of Changes

In all subsequent handlers and service business logic implementation, the `find by id` for the view has been replaced by the `find by id and status code` for the event:

- cancelTransaction
- requestAuthorizationTransaction
- addUserReceipt

This replace is necessary in order to use only event collection to get and check transaction data, since the view update could happen with some delay. Moreover, the only truth source of ecommerce is the events' collection and not the view, so checking transaction data from the view is formally wrong.

Also all the accessory methods invoked by that services and handlers that apply its own business logic to the document view, now use the transaction activated event document.

#### Motivation and Context

The ecommerce cdc introduction makes the view update asynchrounos, so all the payment process api that rely on the view checking need to be update, relying on the event data checking. 
The only one api that relies on the view is the outcome one, and it isn't object of this PR.

All Junit tests have been updatedand deep tests have been performed on dev env. 
All integration and api tests continue to work right.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/pagopa/transactions/projections/handlers/v2/AuthorizationRequestProjectionHandler.java
+++ b/src/main/java/it/pagopa/transactions/projections/handlers/v2/AuthorizationRequestProjectionHandler.java
@@ -41,17 +41,25 @@ public class AuthorizationRequestProjectionHandler
     public Mono<Transaction> handle(AuthorizationRequestedEventData authorizationRequestedEventData) {
         AuthorizationRequestData data = authorizationRequestedEventData.authorizationRequestData();
         String creationDate = authorizationRequestedEventData.event().getCreationDate();
-        return transactionsViewRepository.findById(data.transactionId().value())
-                .cast(Transaction.class)
-                .switchIfEmpty(
-                        Mono.error(
-                                new TransactionNotFoundException(
-                                        data.transactionId().value()
-                                )
-                        )
-                )
+        return Mono.just(transactionsviewUpdateEnabled)
+                .filter(t -> t)
                 .flatMap(
-                        transactionDocument -> conditionallySaveTransactionView(transactionDocument, data, creationDate)
+                        t -> transactionsViewRepository.findById(data.transactionId().value())
+                                .cast(Transaction.class)
+                                .switchIfEmpty(
+                                        Mono.error(
+                                                new TransactionNotFoundException(
+                                                        data.transactionId().value()
+                                                )
+                                        )
+                                )
+                                .flatMap(
+                                        transactionDocument -> conditionallySaveTransactionView(
+                                                transactionDocument,
+                                                data,
+                                                creationDate
+                                        )
+                                )
                 );
     }
 
--- a/src/test/java/it/pagopa/transactions/projections/handlers/v2/AuthorizationRequestProjectionHandlerTest.java
+++ b/src/test/java/it/pagopa/transactions/projections/handlers/v2/AuthorizationRequestProjectionHandlerTest.java
@@ -151,18 +151,11 @@ class AuthorizationRequestProjectionHandlerTest {
                 false
         );
 
-        when(transactionsViewRepository.findById(initialDocument.getTransactionId()))
-                .thenReturn(Mono.just(initialDocument));
-
         StepVerifier.create(authorizationRequestProjectionHandler.handle(authorizationData))
-                .expectNextMatches(
-                        transaction -> transaction.getStatus().equals(TransactionStatusDto.AUTHORIZATION_REQUESTED) &&
-                                transaction.getFeeTotal().equals(fee) &&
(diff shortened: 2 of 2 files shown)


src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionUserCancelHandler.java
```diff
@@ -48,6 +50,13 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionUserCancelCommand command
         return transaction
                 .filter(tx -> tx.getStatus().equals(TransactionStatusDto.ACTIVATED))
                 .switchIfEmpty(Mono.error(new AlreadyProcessedException(command.getData())))
+                .cast(TransactionActivated.class)
+                .filter(
+                        tx -> ((command.getXUserId() == null && tx.getTransactionActivatedData().getUserId() == null) ||
+                                (command.getXUserId() != null && command.getXUserId().toString()
+                                        .equals(tx.getTransactionActivatedData().getUserId())))
+                )
```
pietro-tota, line 58:
```suggestion
                .filter(Objects.equals(Objects.toString(command.getXUserId())),tx.getTransactionActivatedData().getUserId())
```

this condition can be simplified using java.utils.Objects.equals method that return true if both input objects are null or equals be meaning of equals method

ciuffagianluca: Sounds good, but the `command.getXUserId()` could be null, and applying this code i have an error since `Object.equals `is static and cannot be used in non-static context

ciuffagianluca: Ok I see that if the `command.getXUserId()` is null it's not a problem

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
```diff
@@ -1457,45 +1456,46 @@ private Mono<Boolean> wasTransactionAuthorized(
...
     @Retry(name = "addUserReceipt")
     public Mono<TransactionInfoDto> addUserReceipt(
-            String transactionId,
-            AddUserReceiptRequestDto addUserReceiptRequest
+                                                   String transactionId,
+                                                   AddUserReceiptRequestDto addUserReceiptRequest
     ) {
-        return transactionsViewRepository
-                .findById(transactionId)
+        return eventsRepository
```
pietro-tota, line 1462: What do you think if, instead of retrieving activated event, we anticipate the reduce operation done into handler here (passing the base transaction event in the handler) using the base transaction domain object to retrieve informations such as rpt ids etc.. this way we can save a select on event store (that is, reduce operation retrieve the same event retrieved here).
This can be applied to all methods, also to authorization requested, avoiding same query to be performed twice

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionUserCancelHandler.java
```diff
@@ -48,6 +51,14 @@ public Mono<BaseTransactionEvent<?>> handle(TransactionUserCancelCommand command
         return transaction
                 .filter(tx -> tx.getStatus().equals(TransactionStatusDto.ACTIVATED))
                 .switchIfEmpty(Mono.error(new AlreadyProcessedException(command.getData())))
+                .cast(TransactionActivated.class)
+                .filter(
+                        tx -> Objects.equals(
+                                Objects.toString(command.getXUserId()),
```
infantesimone, line 57: Why not compare the UUID directly, avoiding the conversion to strings?

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
infantesimone: Good point! I'm a bit concerned about the size of the PR, maybe it's better to create a separate optimization task dedicated to this aspect?

infantesimone (approved):

CianoDanilo (dismissed):

src/main/java/it/pagopa/transactions/services/v1/TransactionsService.java
pietro-tota: Ok for me, transaction activated event and view are almost the same in size so there would not be too much difference in db load in terms of consumed RU's with this implementation.

"Reduce operation" anticipation can be posticipate to a separate PR :D

sfarralorenzo (dismissed):

infantesimone (approved):

PR #752, opened 2026-05-07
Chk 4810 lock post auth request only first payment token

ciuffagianluca:
Depends on https://github.com/pagopa/pagopa-ecommerce-local/pull/259

Add lock for POST auth-request for payment-token. This change prevent multiple and parallel authorization request versus the payment gateway for the same payment-token but different transactionId. 

#### List of Changes

Added a lock for  POST auth-request based on the payment token. It lasts for 20 seconds and prevent a possibile double authorization on the same payment token and different transactionId (for multiple parallel payment activations). 

#### Motivation and Context

prevent multiple authoriazion for the same payment token

- JUnit test with payment token locking and payment token sorted locking
- Api test dedicated to this specific case

#### Types of changes

#### Checklist:


.devops/code-review-pipelines.yml
```diff
@@ -15,7 +15,7 @@ resources:
     - repository: pagopaEcommerceLocal
       type: github
       name: pagopa/pagopa-ecommerce-local
-      ref: main
+      ref: CHK-4810-introduce-post-auth-request-payment-token-lock
```
ciuffagianluca:
```suggestion
      ref: main
```

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestAuthorizationHandler.java
```diff
@@ -204,7 +213,7 @@ public Mono<Tuple2<RequestAuthorizationResponseDto, TransactionAuthorizationRequ
                 })
                 .flatMap(t -> {
                     TransactionId transactionId = t.getTransactionId();
-                    ExclusiveLockDocument lockDocument = new ExclusiveLockDocument(
+                    ExclusiveLockDocument transactioIdLockDocument = new ExclusiveLockDocument(
```
pietro-tota:
```suggestion
                    ExclusiveLockDocument transactionIdLockDocument = new ExclusiveLockDocument(
```

please add missing `n`

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestAuthorizationHandler.java
```diff
@@ -214,18 +223,43 @@ public Mono<Tuple2<RequestAuthorizationResponseDto, TransactionAuthorizationRequ
...
                             .flatMap(lockAcquired -> {
                                 if (Boolean.FALSE.equals(lockAcquired)) {
-                                    return Mono.error(new LockNotAcquiredException(transactionId, lockDocument));
+                                    return Mono.error(new LockNotAcquiredException(transactionId, transactioIdLockDocument));
+                                }
+                                return Mono.just(t);
+                            });
+                })
+                .flatMap(t -> {
+                            String firstPaymentToken = t.getTransactionActivatedData().getPaymentNotices().stream().map(PaymentNotice::getPaymentToken).findFirst().orElse(null);
```
pietro-tota:
```suggestion
                            String firstPaymentToken = t.getTransactionActivatedData().getPaymentNotices().stream().map(PaymentNotice::getPaymentToken).sorted().findFirst().orElseThrow();
```

1. take the first of all payment token sorted alphabetically, this will protect us in case an EC starts two carts requests with payment notices in different orders
2. do not use orElse(null) default supplier, prefer throw exception here since we cannot have a transaction authorized without at least one payment token in the paymentTokenList, so if it will never happen -> code side it should be an error instead of a null returned value, causing api to fail in the exact point where we expect to find the payment token to be used for lock

pietro-tota (dismissed):
Well done, just some trivial issue

CianoDanilo (dismissed):

src/main/java/it/pagopa/transactions/commands/handlers/v2/TransactionRequestAuthorizationHandler.java
ciuffagianluca: fixed [here](4f33a356679ce282280865a48eefcc5167e16a27)

pietro-tota (dismissed):

pietro-tota (approved):

CianoDanilo (approved):

PR #758, opened 2026-06-03
chore(spring-profiler): [CHK-4847] add human readable profiler for local env

sterondinella:
#### List of Changes
- Add human readable format profiler for local env
- Add profile setting in application.properties-test, needed for testing log

#### Motivation and Context
The current json format profiler for log is not human readable and not usefull for local develop.
This human readable profile for local env will help to have a better develop experience.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/test/resources/application-tests.properties
+++ b/src/test/resources/application-tests.properties
@@ -127,3 +127,4 @@ ecommercePaymentMethodsHandler.connectionTimeout=10000
 ecommercePaymentMethodsHandler.apiKey=payment-method-handler-api-key
 ecommercePaymentMethodsHandler.enabled=false
 closureRequestedRetryDelay.visibilityTimeoutSeconds=20
+spring.profiles.active=k8s
--- a/helm/values-dev.yaml
+++ b/helm/values-dev.yaml
@@ -97,6 +97,7 @@ microservice-chart:
           metricAggregationType: Count
           targetValue: "30"
   envConfig:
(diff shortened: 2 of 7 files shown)


ivanferrando-pagopa (approved):

ivanferrando-pagopa:
I think you can remove the `chart-minor`, `minor` and `patch` labels and keep only the `skip-release` label

src/test/resources/application-tests.properties
```diff
@@ -127,3 +127,4 @@ ecommercePaymentMethodsHandler.connectionTimeout=10000
 ecommercePaymentMethodsHandler.apiKey=payment-method-handler-api-key
 ecommercePaymentMethodsHandler.enabled=false
 closureRequestedRetryDelay.visibilityTimeoutSeconds=20
+spring.profiles.active=k8s
```
CianoDanilo, line 130: Shouldn't we use local profile for tests?

sterondinella: Using local profile break the [LogFormatterTests](https://github.com/pagopa/pagopa-ecommerce-transactions-service/blob/main/src/test/java/it/pagopa/transactions/utils/LogFormatterTests.java), I tried to force the use of the _k8s_ profile during the testing with the `@ActiveProfiles` annotation but the logger ignore it.

CianoDanilo (approved):

PR #762, opened 2026-06-10
feat: [PIDM-2073] - make redirect url mapping configuration dynamic

pietro-tota:
#### List of Changes
Make redirect url mapping configuration dynamic

#### Motivation and Context
Make redirect url mapping configuration dynamic to support newly added matching parameters such as pspChannelId. this have been done in order to support postepay psp merge

#### Types of changes

#### Checklist:


infantesimone (dismissed):

FrancescoDiTonnoJakala (dismissed):

CianoDanilo (dismissed):

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
```diff
@@ -526,19 +529,26 @@ public Mono<RedirectUrlResponseDto> requestRedirectUrlAuthorization(
...
+                                    .getRedirectUrlForCriteria(
+                                            Map.of(
+                                                    RedirectUrlMappingCriteria.TOUCHPOINT,
+                                                    touchpoint.name(),
+                                                    RedirectUrlMappingCriteria.PSP_ID,
+                                                    authorizationData.pspId(),
+                                                    RedirectUrlMappingCriteria.PAYMENT_TYPE_CODE,
+                                                    authorizationData.paymentTypeCode(),
+                                                    RedirectUrlMappingCriteria.PSP_CHANNEL_ID,
+                                                    authorizationData.pspChannelCode()
```
dylantangredi-jakala: I see from commons that pspChannelCode does not have the NotNull annotation https://github.com/pagopa/pagopa-ecommerce-commons/blob/99beab5695528f119424f44c39918dbbd5d9253c/src/main/java/it/pagopa/ecommerce/commons/documents/v2/TransactionAuthorizationRequestData.java#L26

if Map.of rejects null values could this throw a null pointer? Or the pspChannelCode can never be null in any case?

(I think the old code never used this lookup so it was never a potential issue?)

README.md
```diff
@@ -78,10 +78,10 @@ These are all environment variables needed by the application:
...
 | SESSION_URL_NOTIFICATION_URL                                 |      | Url used into npg order build request to enhance the notificationUrl field                                                                                                      | string  |         |
 | NODE_FORWARDER_API_KEY                                       |      | Node forwarder api key                                                                                                                                                          | string  |         |
-| REDIRECT_PAYMENT_TYPE_CODE_LIST                              |      | List of all redirect payment type codes that are expected to be present in other redirect configurations such as REDIRECT_URL_MAPPING (used for configuration cross validation) | string  |         |
+| REDIRECT_EXPECTED_MATCHING_CRITERIA                              |      | List of all redirect payment type codes that are expected to be present in other redirect configurations such as REDIRECT_URL_CONFIGURATION (used for configuration cross validation) | string  |         |
```
dylantangredi-jakala: shoould we update the description too? from type codes to matching criteria since now its a json criteria array

src/main/java/it/pagopa/transactions/configurations/RedirectConfigurationsBuilder.java
```diff
@@ -17,26 +14,26 @@
...
      */
     @Bean
-    public RedirectKeysConfiguration redirectKeysConfig(
-                                                        @Value(
-                                                            "${redirect.paymentTypeCodeList}"
-                                                        ) Set<String> paymentTypeCodeList,
-                                                        @Value(
-                                                            "#{${redirect.pspUrlMapping}}"
-                                                        ) Map<String, String> pspUrlMapping
+    public RedirectUrlMappingConf redirectKeysConfig(
```
dylantangredi-jakala: should we rename this for consistency since the return type is changed? w
on the related PR on the evnet dispatcher is called `redirectBeApiCallUriConf ` for example

dylantangredi-jakala (commented):
LGTM, only a doubt and some comsetics

src/main/java/it/pagopa/transactions/client/PaymentGatewayClient.java
pietro-tota: good catch: pspChannelCode should never be null since it's the Node channel used as configured but since it's retrieved from GEC and its nullable in openapi I would fix it to handle this as null value as well

pietro-tota: [fixed with https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/762/commits](https://github.com/pagopa/pagopa-ecommerce-transactions-service/pull/762/changes/6916ea16b1a9e5e007ca4f86835e078a50fa313c) (parameter should always be present, the "null" value will be used in case of null one).

If configuration have channelId parameter (the one configured in secrets) an exact match will be attempt, otherwise the parameter will be ignored

dylantangredi-jakala (approved):
Good work!

